PR #9, opened 2018-02-28
new script to aggregate some Mapillary labels

garaud:
* resizing
* cropping
* group the 66 differents labls into 11 labels

### DIFF
--- a/sources/aggregate_label.py
+++ b/sources/aggregate_label.py
@@ -1,0 +1,155 @@
+"""Image resizing and aggregation labels
+
+Can pass this script with the GNU command 'parallel' in order to process the
+images in parallel.
+
+cat image.list | parallel "python aggregate_label.py -s=256 -k={}"
+"""
+
+import os
+import json
+
+import numpy as np
+import pandas as pd
+
+from PIL import Image
+
+import utils
+
+
+DATASET = 'mapillary'
+
+
+def _split_object_label(df):
+    """Split 'object' label into three sub-labels
+
+    - object
+    - vehicle
+    - traffic
+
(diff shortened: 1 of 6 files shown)


garaud:
I should also carry out the same processing to validation images.

Maybe we can refactor the `populate` function of the mapillary `Dataset` class in order to carry out the following things:

* resize & croppping
* group labels (or not)
* save new train (and label) images
* store the label occurrence (groupped or not) into a JSON file

delhomer:
That's OK for this module, it works fine taken separately. The idea looks great if we can simplify and group labels (a lot of them are rare, and consequently, hard to detect), as well as if we can accelerate the dataset populating (as an example: more than 45 minutes for a 512*512 training+validation dataset).

Two questions, though:
- Could it be possible to integrate it into the project pipeline, i.e. as a pre-requisite of training a neural network?
- Have you looked at the label images? It seems that they are in a black&white mode, instead of considering Mapillary glossary colors.

garaud:
* modify the aggregate labels script: no more image resizing. Just write a new JSON configuration file with the suffix `aggregate`
* the `self.image_info[0]['labels']` is now a dict with the label ids and the occurrence as a value (instead of a list previously)
* new `-al/--aggregate` option for the train.py script

Ready to proofreading and proof-launching

Note that the structure of the JSON config file (aggregated or not) has been slightly modified. It's better to generate new ones.

delhomer:
Some changes done around glossary printing and testing features. A particular focus has been done on label structure transformation (from list to dictionary) in dataset objects. This was especially source of bugs for shape dataset.

At this time, it seems reasonable to merge. If further issues arise, an additional branch will have to be opened (for example, regarding label gathering process).

PR #11, opened 2018-03-07
Inference

delhomer:
Add a test.py module to do inference with a trained model. The module is callable (just do python3 test.py [args]), it applies the test() method within the ConvolutionalNeuralNetwork class.
In this method, the trained model is recovered (test.py arguments must correspond to arguments previously passed to train.py) and it is applied on testing images. Results are stored into a `.json` file, under `datarep/results/name/` repository (the filename depends on the training step).

### DIFF
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -132,7 +132,7 @@ class Dataset(object):
                                      "category": category,
                                      "color": color}
 
-    def populate(self, datadir, nb_images=None):
+    def populate(self, datadir, nb_images=None, labelling=True):
         """ Populate the dataset with images contained into `datadir` directory
  
        Parameter:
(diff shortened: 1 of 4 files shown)


delhomer:
As an enhancement, we could imagine that each training produces a configuration file that will have to be used in a further inference step. This step seems useful as it is currently hard to remind which arguments to pass to test.py, when exploiting a trained model.

delhomer:
No configuration file; however the instance name is used to recover the model characteristics (image size, network size). From now, model graph and checkpoints are stored as `<instance-name>_<image-size>_<network-size>`. A model will be tested by applying this whole format to `-n` parameter of test.py.

delhomer:
OK for shape dataset, KO for Mapillary dataset -> manage case where they are no testing labels (the definition of an inference set, actually).

PR #15, opened 2018-03-13
add a Python script for sematic segmentation with Keras

garaud:
ready to review

### DIFF
--- a/sources/segnet_keras.py
+++ b/sources/segnet_keras.py
@@ -1,0 +1,216 @@
+"""A first draft neural network with Keras for semantic segmentation (Mapillary)
+
+This script uses the aggregated labels of Mapillary, i.e. 66 classes turn into 11 classes.
+"""
+
+import os
+import json
+import os.path as osp
+from datetime import datetime
+
+import numpy as np
+
+import keras
+from keras.preprocessing.image import ImageDataGenerator
+from keras.layers import (Input, Conv2D, MaxPool2D, Conv2DTranspose,
+                          Activation, BatchNormalization)
+from keras.models import Model
+
+
+_now = datetime.now()
+SIZE = 256
+BATCH_SIZE = 32
+# number of images per epoch
+NB_IMAGES = 18000
+NB_VALIDATION_IMAGES = 2000
+STEPS = NB_IMAGES // BATCH_SIZE
+print("STEP {}".format(STEPS))
+VALIDATION_STEPS = NB_VALIDATION_IMAGES // BATCH_SIZE
+print("VALIDATION STEPS {}".format(VALIDATION_STEPS))
+CLASSES = 11
+EPOCHS = 30
+DATADIR = "../data/mapillary"
+OUTPUT_DIR = osp.join(DATADIR, 'segnet-output', 'run-{}'.format(_now.strftime("%Y-%m-%dT%H:%M")))
+OUTPUT_DIR = OUTPUT_DIR
+CONFIG = osp.join(DATADIR, 'config_aggregate.json')
+TRAINING = osp.join(DATADIR, "training_aggregate_{}".format(SIZE))
+VALIDATION = osp.join(DATADIR, "validation_aggregate_{}".format(SIZE))
+SEED = 1337
+
+
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+
+def to_categorical(img, label_ids):
+    """One-hot encoder for a labelled image
+
+    Parameters
+    ----------
+    img : ndarray
+        2D or 3D (if batched)
+
+    Returns
+    -------
+    ndarray
+        Occurrence of the ith label for each pixel
+    """
+    img = img.squeeze()
+    input_shape = img.shape
+    img = img.ravel().astype(np.uint8)
+    for idx, label in enumerate(label_ids):
+        mask = img == label
+        img[mask] = idx
+    n = img.shape[0]
+    categorical = np.zeros((n, len(label_ids)))
+    categorical[np.arange(n), img] = 1
+    output_shape = input_shape + (len(label_ids), )
+    return categorical.reshape(output_shape)
+
+
+def block_encoder(x, num_filters, kernel_size=3):
+    """Encoder block
+
+    Convolution + Batch normalization + ReLU activation + Max Pooling
+
+    Parameters
+    ----------
+    x : keras.Layer
+    num_filters : int
+    kernel_size : int (3 by default)
+
+    Returns
+    -------
+    keras.Layer
+    """
+    x = Conv2D(num_filters, kernel_size=kernel_size, padding='same')(x)
+    x = BatchNormalization()(x)
+    x = Activation('relu')(x)
+    x = MaxPool2D(2)(x)
+    return x
+
+
+def block_decoder(x, num_filters, kernel_size=3):
+    """Decoder block.
+
+    For now, just a Convolution Transpose
+
+    Parameters
+    ----------
+    x : keras.Layer
+    num_filters : int
+    kernel_size : int (3 by default)
+
+    Returns
+    -------
+    keras.Layer
+    """
+    x = Conv2DTranspose(num_filters, padding='same', activation='relu',
+                        kernel_size=kernel_size, strides=2)(x)
+    return x
+
+
+def read_config():
+    """Juste read the JSON configuration file.
+    """
+    with open(CONFIG) as fobj:
+        return json.load(fobj)
+
+
+config = read_config()
+labels = config['labels']
+label_ids = np.array([x['id'] for x in labels])
+
+train_datagen = ImageDataGenerator()
+test_datagen = ImageDataGenerator()
+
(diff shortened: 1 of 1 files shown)


sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+NB_IMAGES = 18000
+NB_VALIDATION_IMAGES = 2000
+STEPS = NB_IMAGES // BATCH_SIZE
+print("STEP {}".format(STEPS))
+VALIDATION_STEPS = NB_VALIDATION_IMAGES // BATCH_SIZE
+print("VALIDATION STEPS {}".format(VALIDATION_STEPS))
+CLASSES = 11
+EPOCHS = 30
+DATADIR = "../data/mapillary"
+OUTPUT_DIR = osp.join(DATADIR, 'segnet-output', 'run-{}'.format(_now.strftime("%Y-%m-%dT%H:%M")))
```
delhomer, line 33: Could it be possible to standardize the output repository on the model of feature detection output? (Or at least, to standardize the output repositories to whatever format, so as feature detection and semantic segmentation gives comparable output repos.)

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+OUTPUT_DIR = OUTPUT_DIR
```
delhomer, line 34: Useless line.

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+CONFIG = osp.join(DATADIR, 'config_aggregate.json')
+TRAINING = osp.join(DATADIR, "training_aggregate_{}".format(SIZE))
+VALIDATION = osp.join(DATADIR, "validation_aggregate_{}".format(SIZE))
+SEED = 1337
+
```
delhomer, line 39: More generally on this part, I prefer considering a main module with command arguments, so as to use the program on different configurations.

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+    input_shape = img.shape
+    img = img.ravel().astype(np.uint8)
+    for idx, label in enumerate(label_ids):
+        mask = img == label
+        img[mask] = idx
+    n = img.shape[0]
+    categorical = np.zeros((n, len(label_ids)))
+    categorical[np.arange(n), img] = 1
+    output_shape = input_shape + (len(label_ids), )
+    return categorical.reshape(output_shape)
```
delhomer, line 67: Very nice!

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Returns
+    -------
+    keras.Layer
+    """
+    x = Conv2D(num_filters, kernel_size=kernel_size, padding='same')(x)
+    x = BatchNormalization()(x)
+    x = Activation('relu')(x)
+    x = MaxPool2D(2)(x)
+    return x
```
delhomer, line 89: I like this way of doing, *i.e.* designing smaller neural network components, that looks common in the state of the art. Just one little comment here about the pooling operation: maybe we will have to consider blocks with and blocks without pooling, as we will need deeper networks. This should become parameterizable on a further release (not so urgent, anyway).

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def block_decoder(x, num_filters, kernel_size=3):
+    """Decoder block.
+
+    For now, just a Convolution Transpose
```
delhomer, line 95: See my previous comment, we agree on the diagnosis. :)

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+    Returns
+    -------
+    keras.Layer
+    """
+    x = Conv2DTranspose(num_filters, padding='same', activation='relu',
+                        kernel_size=kernel_size, strides=2)(x)
+    return x
+
+
+def read_config():
```
delhomer, line 112: This function makes me think about the Dataset class designed for feature detection problem. There is probably some duplicated code.

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+    batch_size=BATCH_SIZE,
+    color_mode='grayscale',
+    class_mode=None,
+    seed=SEED)
+masks_generator = iter(to_categorical(x, label_ids) for x in masks_generator)
+train_generator = zip(images_generator, masks_generator)
+validation_mask_generator = iter(to_categorical(x, label_ids)
+                                 for x in validation_mask_generator)
+validation_generator = zip(validation_images_generator, validation_mask_generator)
+
```
delhomer, line 161: All this part (from line 123 to 161) is a Keras major benefit: it is shorter, clearer, and easier to code. It is more understandable to introduce our own dataset on this way, I have to point it out.

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+# Number of feature maps equals to the number of classes, i.e. labels
+x = Conv2DTranspose(CLASSES, kernel_size=2, padding='same')(x)
+
+# Last layer
+y = Activation('softmax', name='target')(x)
+
+model = Model(img_input, y)
+model.compile(loss='categorical_crossentropy',
+              optimizer='adam',
+              metrics=['acc', 'mae'])
```
delhomer, line 185: I'm wondering what sort of metrics we may consider here. Maybe it is possible to add some hand-made ones? We will have to consider IoU in semantic segmentation, either in `loss` or in `metrics` arguments of this function. That will be an important challenge here.

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+json_model = model.to_json()
+with open(osp.join(OUTPUT_DIR, "segnet-model-keras.json"), "w") as fobj:
```
delhomer, line 187: To integrate in a clean output repository (see first comment).

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+    verbose=0,
+    save_best_only=True,
+    save_weights_only=False,
+    mode='auto', period=1)
+earlystop = keras.callbacks.EarlyStopping(monitor='val_acc',
+                                          min_delta=0.002,
+                                          patience=10,
+                                          verbose=0,
+                                          mode='auto')
+hist = model.fit_generator(train_generator, epochs=EPOCHS,
```
delhomer, line 204: Second huge benefit of Keras: the callbacks, it looks very pythonic. :) (And less headache-maker than TensorFlow way of doing.)

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+                           callbacks=[tsboard, check, earlystop],
+                           steps_per_epoch=STEPS,
+                           validation_data=validation_generator,
+                           validation_steps=VALIDATION_STEPS)
+metrics = {"epoch": hist.epoch,
+           "metrics": hist.history,
+           "params": hist.params}
```
delhomer, line 211: Here, "metrics" and "params" come necessarily from the compilation parametrization?

sources/segnet_keras.py
```diff
@@ -0,0 +1,216 @@
...
+with open(osp.join(OUTPUT_DIR, "hist-metrics.json"), "w") as fobj:
+    json.dump(metrics, fobj)
+fname = "final-segnet-batch-{}-size-{}-max-epoch-{}.h5"
+fname = fname.format(BATCH_SIZE, SIZE, EPOCHS)
+model.save(osp.join(OUTPUT_DIR, fname))
```
delhomer, line 216: Same remark as before regarding output backup. Moreover what's the difference between this `h5` file and checkpoint files? Same nature, but different training state ? Or the final segnet result includes more features ?

delhomer (commented):
Nice first draw for doing semantic segmentation!
Several remarks, however, before to integrate it into the project:
- I believe we should standardize the code (decomposition with train.py, model.py and dataset.py, like in feature detection problem) to make it even clearer that it is already. It should allow us to generalize it for studying different cases (and thus, consider parameters in a main-method manner).
- Standardization of the output repo as well. The feature detection results are stored as if they are the only one to consider, however if we add the semantic segmentation results, we should do like:
feature_detection
    |_ log
    |_ checkpoints
semantic_segmentation
    |_ log
    |_ checkpoints

I'm of course available to take part to the changes!

sources/segnet_keras.py
garaud: Yes, this is a good idea.

sources/segnet_keras.py
garaud: Agreed to add some custom metrics. I tihink the implementation of the IoU can be occur in its specific issue. See #21

sources/segnet_keras.py
garaud: Yes, it's the same nature. Just the final one :)

sources/segnet_keras.py
garaud: It sounds me good.

Issue #19, opened 2018-03-14
Introduce semantic segmentation in the project


delhomer:
A more ambitious kind of image detection problem is semantic segmentation, *i.e.* the association of each pixel with a class of the glossary, so as to get a fully-labeled picture (see [Mapillary dataset](https://www.mapillary.com/dataset/vistas?lat=20&lng=0&z=1.5), and some examples in this [repo](https://github.com/Oslandia/streetscene_CNN/tree/master/images)).

Other state-of-the-art models exist in this way, one can refer for instance to [SEC](https://github.com/kolesman/SEC) or [Mask-RCNN](https://github.com/matterport/Mask_RCNN).

garaud:
Can this issue be closed with the recent pull request #15 ? Or can you prefer waiting for the "true" semantic segmentation integration in the `models.py` module?

delhomer:
We may close this one, as it is related to the "introduction" of semantic segmentation. And re-open another (others) issues progressively if needed.

PR #24, opened 2018-03-20
I/O repos

### LINKED ISSUE #23: Thoughts about the input/ouput folders structure and design experiment
I think we should separate the input and output folders which can share the same root directory, e.g. `root_dir = '../data/streetscene'`.

* input folder with the raw datasets
  - `input/mapillary` where you can find the `tranining` and `validation` folders and the further preprocess folders (resized and/or aggregated images)
  - `input/shapes`

* output folder with the different output result for the different dataset
  - `output/mapillary`
  - `output/shapes`

and create a folder for each run. A number of parameter names and the date can occur in the folder name. A meta.json or YAML file inside this folder which summarizes the list of input parameters and hyptohesis.

delhomer:
Modify the structure of input/output repositories. The new architecture is described in issue #23 .

The goal of this PR is to distinguished each instance (instance name, image size, network size, label aggregation, dropout rate, learning rate) when creating model backups, as well as to distinguish clearly between research problems (for instance feature detection is the only problem addressed in `master`, however semantic segmentation will be merged soon, see #19 ). The PR sorts as clearly as possible preprocessed data too, in a dedicated directory : a preprocessed data set is essentially characterized by image size and label aggregation.

Fix #23.

### DIFF
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -186,11 +186,11 @@ class Dataset(object):
                 a[mask] = root_id
         return Image.fromarray(a, mode=image.mode)
 
-    def _preprocess(self, datadir, image_filename, aggregate, labelling=True):
+    def _preprocess(self, image_filename, output_dir, aggregate, labelling=True):
         """Resize/crop then save the training & label images
 
-        :param datadir: str
         :param image_filaname: str
+        :param output_dir: str
         :param aggregate: boolean
         :param labelling: boolean
         :return: dict - Key/values with the filenames and label ids
@@ -206,7 +206,7 @@ class Dataset(object):
         final_img_in = utils.mono_crop_image(img_in, crop_pix)
 
         # save final image
-        new_in_filename = os.path.join(datadir, 'images', image_filename.split('/')[-1])
+        new_in_filename = os.path.join(output_dir, 'images', image_filename.split('/')[-1])
         final_img_in.save(new_in_filename)
 
         # label_filename vs label image
@@ -221,7 +221,7 @@ class Dataset(object):
                 final_img_out = self.group_image_label(final_img_out)
 
             labels = utils.mapillary_label_building(final_img_out, self.label_ids)
-            new_out_filename = os.path.join(datadir, 'labels', label_filename.split('/')[-1])
+            new_out_filename = os.path.join(output_dir, 'labels', label_filename.split('/')[-1])
             final_img_out.save(new_out_filename)
         else:
             new_out_filename = None
@@ -232,14 +232,15 @@ class Dataset(object):
                 "label_filename": new_out_filename,
                 "labels": labels}
 
-    def populate(self, datadir, nb_images=None, aggregate=False, labelling=True):
+    def populate(self, input_dir, output_dir, nb_images=None, aggregate=False, labelling=True):
         """ Populate the dataset with images contained into `datadir` directory
 
         Parameters
         ----------
-        datadir : object
-            String designing the relative path of the directory that contains
-        new images
+        input_dir : str
+            Path of the directory that contains input images
+        output_dir : str
+            Path of the directory where the preprocessed image must be saved
         nb_images : integer
             Number of images to be considered in the dataset; if None, consider the whole
         repository
@@ -248,13 +249,10 @@ class Dataset(object):
         labelling: boolean
             If True labels are recovered from dataset, otherwise dummy label are generated
         """
-        raw_datadir, folder = os.path.split(datadir)
-        raw_datadir = os.path.join(raw_datadir, folder.split('_')[0])
-        image_dir = os.path.join(raw_datadir, "images")
-        image_list = os.listdir(image_dir)[:nb_images]
-        image_list_longname = [os.path.join(image_dir, l) for l in image_list]
+        image_list = os.listdir(os.path.join(input_dir, "images"))[:nb_images]
+        image_list_longname = [os.path.join(input_dir, "images", l) for l in image_list]
         with Pool() as p:
-            labels = p.starmap(self._preprocess, [(datadir, x, aggregate, labelling) for x in image_list_longname])
+            labels = p.starmap(self._preprocess, [(x, output_dir, aggregate, labelling) for x in image_list_longname])
         self.image_info = {k: v for k,v in enumerate(labels)}
 
     def save(self, filename):
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -232,15 +232,15 @@ class Dataset(object):
                 "label_filename": new_out_filename,
                 "labels": labels}
 
-    def populate(self, input_dir, output_dir, nb_images=None, aggregate=False, labelling=True):
+    def populate(self, output_dir, input_dir, nb_images=None, aggregate=False, labelling=True):
         """ Populate the dataset with images contained into `datadir` directory
 
         Parameters
         ----------
-        input_dir : str
-            Path of the directory that contains input images
         output_dir : str
             Path of the directory where the preprocessed image must be saved
+        input_dir : str
+            Path of the directory that contains input images
         nb_images : integer
             Number of images to be considered in the dataset; if None, consider the whole
         repository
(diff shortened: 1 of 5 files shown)


sources/test.py
```diff
@@ -73,15 +79,27 @@
...
+
     # Data path and repository management
     dataset_repo = os.path.join(args.datapath, args.dataset)
-    testing_name = "testing_" + str(image_size)
-    os.makedirs(os.path.join(dataset_repo, testing_name, "images"), exist_ok=True)
-    testing_filename = os.path.join(dataset_repo, testing_name + '.json')
+    input_repo = os.path.join(dataset_repo, "input")
+    preprocessed_repo = str(image_size) + "_" + aggregate_value
+    preprocessed_path = os.path.join(dataset_repo, "preprocessed", preprocessed_repo)
+    testing_filename = os.path.join(preprocessed_path, "testing.json")
```
garaud: What is the purpose of the `testing.json` file?

sources/train.py
```diff
@@ -150,27 +154,39 @@
...
-    suffix = "" if not args.aggregate_label else "aggregate_"
-    training_name = "training_" + suffix + str(args.image_size)
-    validation_name = "validation_" + suffix + str(args.image_size)
-    os.makedirs(os.path.join(dataset_repo, training_name, 'images'), exist_ok=True)
-    os.makedirs(os.path.join(dataset_repo, training_name, 'labels'), exist_ok=True)
-    os.makedirs(os.path.join(dataset_repo, validation_name, 'images'), exist_ok=True)
-    os.makedirs(os.path.join(dataset_repo, validation_name, 'labels'), exist_ok=True)
-    training_filename = os.path.join(dataset_repo, training_name + '.json')
-    validation_filename = os.path.join(dataset_repo, validation_name + '.json')
+    input_repo = os.path.join(dataset_repo, "input")
```
garaud: Maybe you can put this content into a dedicated function e.g. `prepare_folders` or something else. What do you think?

sources/utils.py
```diff
@@ -377,3 +377,18 @@ def split_list(l, nb_partitions):
     """
     return [l[i:(i+len(l)/nb_partitions)]
             for i in range(0, len(l), int(len(l)/nb_partitions))]
+
+def list_to_str(l):
+    """Transform the input list into a ready-to-print string
+
+    Parameters
+    ----------
+    l: list
```
garaud: Add a whitespace between the argument and its type.

sources/utils.py
```diff
@@ -377,3 +377,18 @@ def split_list(l, nb_partitions):
     """
     return [l[i:(i+len(l)/nb_partitions)]
             for i in range(0, len(l), int(len(l)/nb_partitions))]
+
+def list_to_str(l):
```
garaud: I prefer having a variable name with more than one letter here. I like `seq` for *sequence*. It can be a list, a tuple, etc. It's a collection of strings. Or another name > 1 letter :)

sources/utils.py
```diff
@@ -377,3 +377,18 @@ def split_list(l, nb_partitions):
...
+    ----------
+    l: list
+        Input list that must be transformed
+
+    Returns
+    -------
+    str
+        Printable version of input list
+    """
+    return '-'.join([str(i) for i in l])
```
garaud: You can also avoid the comprehension list and write

```python
'-'.join(str(i) for i in seq)
```

sources/utils.py
garaud: It's really not necessary. You can imagine to have a more generic function if you pass a "separator" parameter, with the value `-` by default.

```python
def list_to_str(seq, sep='-'):
    pass
```

sources/train.py
```diff
@@ -150,27 +154,39 @@
...
+    preprocessed_validation_path = os.path.join(preprocessed_path, "validation")
+    backup_path = os.path.join(dataset_repo, "output", args.model)
+    os.makedirs(os.path.join(preprocessed_training_path, "images"), exist_ok=True)
+    os.makedirs(os.path.join(preprocessed_training_path, "labels"), exist_ok=True)
+    os.makedirs(os.path.join(preprocessed_validation_path, "images"), exist_ok=True)
+    os.makedirs(os.path.join(preprocessed_validation_path, "labels"), exist_ok=True)
+    os.makedirs(backup_path, exist_ok=True)
+
+    # Instance name (name + image size + network size + batch_size + aggregate? + dropout + learning_rate)
+    instance_name = (args.name + "_" + str(args.image_size) + "_" + args.network_size
```
garaud: Maybe you can put all these variables in a list and call the `list_to_str` function with a `_` separator

sources/test.py
```diff
@@ -34,6 +34,8 @@
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description=("Convolutional Neural Netw"
                                                   "ork on street-scene images"))
+    parser.add_argument('-a', '--aggregate-label', action='store_true',
```
garaud, line 37: I think, for further pull requests, we can factorize some code between the `train.py` and `test.py` modules because I see multiple duplication: the parsing of the command line, folders and files checking, etc.

sources/dataset.py
```diff
@@ -186,11 +186,11 @@ def group_image_label(self, image):
                 a[mask] = root_id
         return Image.fromarray(a, mode=image.mode)
 
-    def _preprocess(self, datadir, image_filename, aggregate, labelling=True):
+    def _preprocess(self, image_filename, output_dir, aggregate, labelling=True):
         """Resize/crop then save the training & label images
 
-        :param datadir: str
         :param image_filaname: str
+        :param output_dir: str
```
garaud, line 193: Is is possible to update the docstring format? With numpydoc style?

garaud (changes requested):

sources/test.py
delhomer: The same as `training.json` and `validation.json`: initialize the `Dataset` object that contains all the dataset image characteristics. Even if it is the testing phase, we still have a directory full of thousands of images. And we need to centralize the information so as to read them carefully.

The "labels" key within this file appears as unnecessary, however we need to store dummy labels because of TensorFlow feeding process (labels are expected during image batching, see `define_batch` method in `model.py`). This behavior could certainly be improved, however it seems more accurate to me to open a new dedicated issue.

sources/utils.py
delhomer: Done.

sources/utils.py
delhomer: Done.

sources/utils.py
delhomer: Done.

sources/utils.py
delhomer: The `sep` argument has been added, however I'm not sure to understand your point, when you write "it's really not necessary". Do you mean that this function is useless?

sources/train.py
delhomer: Could be possible. Not sure it is a fundamental improvement, however it would be cleaner, for sure.

sources/test.py
delhomer: Be careful, the command line parsing is not managed exactly in the same manner in both files. The point is that we must point to the accurate `<datadir>/<dataset>/output/<problem>/checkpoints/<instance_name>` checkpoint to recover the model. Either we set all arguments as in `train.py`, or we set only one argument, that is the whole `<instance_name>` string. I chose this second way, as some arguments are not useful in `test.py`. Do you think the first way is better?

sources/dataset.py
delhomer: I suggest to open another issue to address this.

sources/train.py
delhomer: Why not. It will clarify the `train.py` and `test.py` modules. These tedious code lines seems unavoidable though (let's hide them into `utils.py` :) ).

sources/utils.py
garaud: It's not necessary "for now", but I think it's a good design. See a comment above with the list of args and the `_` separator. We already have an good usage for this function with the `sep` parameter!

sources/train.py
garaud: Agreed.

sources/dataset.py
garaud: ok

sources/test.py
garaud: I'm aware that all arguments parsing can not be shared between `train` and `test`. It's just a feeling that some stuff could be factorized when I saw the current diffs. I wonder if it's relevant.

garaud (approved):
Sounds good to me. You can merge.

sources/test.py
delhomer: You're right, it is relevant. The dataset creation and populating, as well as the glossary management, are common. The only parts that must be differentiated are argument parsing and model calling. Do we address it in this PR?

garaud:
No. We can merge as it is.
Raphael Delhome writes:

>
>
>
>> @@ -34,6 +34,8 @@
>  if __name__ == '__main__':
>      parser = argparse.ArgumentParser(description=("Convolutional Neural Netw"
>                                                    "ork on street-scene images"))
> +    parser.add_argument('-a', '--aggregate-label', action='store_true',
>
> You're right, it is relevant. The dataset creation and populating, as well as the glossary management, are common. The only parts that must be differentiated are argument parsing and model calling. Do we address it in this PR?


-- 
Damien Garaud
http://oslandia.com

PR #27, opened 2018-03-22
Dataset

### LINKED ISSUE #20: Update shape dataset definition
The shape dataset has to be extended to labeled images, *i.e.* labeled version of each image must be produced during dataset population, and corresponding file path must be integrated into the dataset structure stored as a `json` file.

This work has still be done for Mapillary dataset.

delhomer:
Define a proper version of Dataset class, considered as a master class with respect to a new MapillaryDataset class, as well as to ShapeDataset. Both classes inherit from Dataset, code has been factorized accordingly.

Additionally, if the dataset does not exist, ShapeDataset population produces new files that correspond to labelled version of images, in order to allow semantic segmentation problem addressing (fix #20 ).

### DIFF
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -49,7 +49,7 @@ class Dataset(object):
             Id of the dataset class that must be returned
         """
         if not class_id in self.class_info.keys():
-            print("Class {} not in the dataset glossary".format(class_id))
+            utils.logger.error("Class {} not in the dataset glossary".format(class_id))
             return None
         return self.class_info[class_id]
 
@@ -62,7 +62,7 @@ class Dataset(object):
             Id of the dataset image that must be returned
         """
         if not image_id in self.image_info.keys():
-            print("Image {} not in the dataset".format(image_id))
+            utils.logger.error("Image {} not in the dataset".format(image_id))
             return None
         return self.image_info[image_id]
 
@@ -104,7 +104,7 @@ class Dataset(object):
         labels = [self.image_info[im]["labels"]
                   for im in self.image_info.keys()]
         if self.get_nb_images() == 0:
-            utils.logger.info("No images in the dataset.")
+            utils.logger.error("No images in the dataset.")
             return None
         else:
             return np.round(np.divide(sum(np.array([list(l.values()) for l in labels])),
@@ -131,7 +131,7 @@ class Dataset(object):
             List of class ids aggregated by the current class_id
         """
         if class_id in self.class_info:
-            print("Class {} already stored into the class set.".format(class_id))
+            utils.logger.error("Class {} already stored into the class set.".format(class_id))
             return None
         self.class_info[class_id] = {"name": class_name,
                                      "category": category,
@@ -210,7 +210,7 @@ class MapillaryDataset(Dataset):
         with open(config_filename) as config_file:
             glossary = json.load(config_file)
         if "labels" not in glossary:
-            print("There is no 'label' key in the provided glossary.")
+            utils.logger.error("There is no 'label' key in the provided glossary.")
             return None
         for lab_id, label in enumerate(glossary["labels"]):
             lab_id = lab_id if 'id' not in label else label['id']
@@ -420,7 +420,7 @@ class ShapeDataset(Dataset):
         number of classes in the dataset
         """
         if image_id in self.image_info.keys():
-            print("Image {} already stored into the class set.".format(image_id))
+            utils.logger.error("Image {} already stored into the class set.".format(image_id))
             return None
         self.image_info[image_id] = {"background": background,
                                      "shape_specs": specifications,
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -338,11 +338,11 @@ class ShapeDataset(Dataset):
         nb_classes : integer
             Number of shape types (either 1, 2 or 3, warning if more)
         """
-        self.add_class(0, "square", ShapeDataset.SQUARE_COLOR, True)
+        self.add_class(0, "square", self.SQUARE_COLOR, True)
         if nb_classes > 1:
-            self.add_class(1, "circle", ShapeDataset.CIRCLE_COLOR, True)
+            self.add_class(1, "circle", self.CIRCLE_COLOR, True)
         if nb_classes > 2:
-            self.add_class(2, "triangle", ShapeDataset.TRIANGLE_COLOR, True)
+            self.add_class(2, "triangle", self.TRIANGLE_COLOR, True)
         if nb_classes > 3:
             utils.logger.warning("Only three classes are considered.")
 
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -33,7 +33,16 @@ import numpy as np
 import utils
 
 class Dataset(object):
-    """
+    """Generic class that describes the behavior of a Dataset object: it is initialized at least
+    with an image size, its class are added always through the same manner, it can be serialized (save) and
+    deserialized (load) from/to a `.json` file
+
+    Attributes
+    ----------
+    image_size : int
+        Size of considered images (height=width), raw images will be resized during the
+    preprocessing
+
     """
     def __init__(self, image_size):
         self.image_size = image_size
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -166,14 +166,14 @@ class Dataset(object):
         """
         with open(filename) as fp:
             ds = json.load(fp)
-            self.image_size = ds["image_size"]
-            self.class_info = {int(k):ds["classes"][k] for k in ds["classes"].keys()}
-            if nb_images is None:
-                self.image_info = {int(k):ds["images"][k] for k in ds["images"].keys()}
-            else:
-                self.image_info = {int(k):ds["images"][k] for k in ds["images"].keys() if int(k) < nb_images}
-            for img_id, info in self.image_info.items():
-                info['labels'] = {int(k): v for k, v in info['labels'].items()}
+        self.image_size = ds["image_size"]
+        self.class_info = {int(k): ds["classes"][k] for k in ds["classes"].keys()}
+        if nb_images is None:
+            self.image_info = {int(k): ds["images"][k] for k in ds["images"].keys()}
+        else:
+            self.image_info = {int(k): ds["images"][k] for k in ds["images"].keys() if int(k) < nb_images}
+        for img_id, info in self.image_info.items():
+            info['labels'] = {int(k): v for k, v in info['labels'].items()}
         utils.logger.info("The dataset has been loaded from {}".format(filename))
 
 class MapillaryDataset(Dataset):
--- a/sources/dataset.py
+++ b/sources/dataset.py
@@ -32,15 +32,12 @@ import numpy as np
 
 import utils
 
-
 class Dataset(object):
-
-    def __init__(self, image_size, glossary_filename):
-        """ Class constructor
-        """
+    """
+    """
+    def __init__(self, image_size):
         self.image_size = image_size
         self.class_info = defaultdict()
-        self.build_glossary(glossary_filename)
         self.image_info = defaultdict()
 
     def get_class(self, class_id):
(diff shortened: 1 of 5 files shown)


sources/dataset.py
```diff
@@ -306,6 +306,10 @@ def populate(self, output_dir, input_dir, nb_images=None, aggregate=False, label
 
 class ShapeDataset(Dataset):
 
+    SQUARE_COLOR = [0, 10, 10]
```
garaud: I think a tuple structure is better because it's immutable.

sources/dataset.py
```diff
@@ -323,11 +327,11 @@ def build_glossary(self, nb_classes):
         nb_classes: integer
             Number of shape types (either 1, 2 or 3, warning if more)
         """
-        self.add_class(0, "square", [0, 10, 10], True)
+        self.add_class(0, "square", ShapeDataset.SQUARE_COLOR, True)
```
garaud: You can write `self.SQUARE_COLOR` because it's a class attribute.

sources/dataset.py
```diff
@@ -426,16 +430,19 @@ def draw_image(self, image_id, datapath):
 
         image = np.ones([self.image_size, self.image_size, 3], dtype=np.uint8)
         image = image * np.array(image_info["background"], dtype=np.uint8)
+        labelled_image = np.ones([self.image_size, self.image_size, 3], dtype=np.uint8) * 255
 
         # Get the center x, y and the size s
         if image_info["labels"][0]:
```
garaud: Just a question: what is the value returned by `image_info["labels"][0]`? Is it a boolean, a digit (0 or 1)? I will be more explicit if you can read something like `if image_info["labels"]["type"] == "circle"`. In this way, you know the type of the geometry shape.

sources/dataset.py
```diff
@@ -426,16 +430,19 @@ def draw_image(self, image_id, datapath):
...
             color, x, y, s = image_info["shape_specs"][0]
             color = tuple(map(int, color))
             image = cv2.rectangle(image, (x - s, y - s), (x + s, y + s), color, -1)
+            labelled_image = cv2.rectangle(labelled_image, (x - s, y - s), (x + s, y + s), self.class_info[0]["color"], -1)
         if image_info["labels"][1]:
             color, x, y, s = image_info["shape_specs"][1]
             color = tuple(map(int, color))
```
garaud: Is `color` stored as string? I wonder why you have to cast the values of `color` to int.

sources/dataset.py
```diff
@@ -32,15 +32,12 @@
 
 import utils
 
-
 class Dataset(object):
-
-    def __init__(self, image_size, glossary_filename):
-        """ Class constructor
-        """
+    """
```
garaud: Maybe you can add some content to the docstring of the class.

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
                                      "aggregate": aggregate,
                                      "color": color}
 
+    def save(self, filename):
+        """Save dataset in a json file indicated by `filename`
+
+        Parameter
```
garaud: Can you add a `s` to `Parameter`?

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        ---------
+        filename: object
```
garaud: There is a whitespace between `filename`and the `:` (`nb_images` too).

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        with open(filename, 'w') as fp:
+            json.dump({"image_size": self.image_size,
+                       "classes": self.class_info,
+                       "images": self.image_info}, fp)
+        utils.logger.info("The dataset has been saved into {}".format(filename))
+
+    def load(self, filename, nb_images=None):
+        """Load a dataset from a json file indicated by `filename`
+
+        Parameter
```
garaud: Same comment about the docstring.

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        ---------
+        filename: object
+            String designing the relative path from where the dataset must be
+        loaded
+        nb_images: integer
+            Number of images that must be loaded (if None, the whole dataset is loaded)
+        """
+        with open(filename) as fp:
+            ds = json.load(fp)
+            self.image_size = ds["image_size"]
```
garaud: You don't have to indent all code in the context manager. You can just open and read the JSON content and go back to a "normal" indentation.

sources/dataset.py
```diff
@@ -32,15 +32,12 @@
...
 class Dataset(object):
-
-    def __init__(self, image_size, glossary_filename):
-        """ Class constructor
-        """
+    """
+    """
+    def __init__(self, image_size):
         self.image_size = image_size
         self.class_info = defaultdict()
```
garaud: I don't understand the usage of two defaultdict for `class_info` and `image_info` attributes. Classic `dicts` such as `self.class_info = {}` are not enough?

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        loaded
+        nb_images: integer
+            Number of images that must be loaded (if None, the whole dataset is loaded)
+        """
+        with open(filename) as fp:
+            ds = json.load(fp)
+            self.image_size = ds["image_size"]
+            self.class_info = {int(k):ds["classes"][k] for k in ds["classes"].keys()}
+            if nb_images is None:
+                self.image_info = {int(k):ds["images"][k] for k in ds["images"].keys()}
```
garaud: You can iterate on a dict without the `.keys()`. By default, when you iterate on a dict, you iterate on its keys.

```python
d = {"a": 12, "b": -7}
[x for x in d]
# equivalent result and better than
[x for x in d.keys()]
```

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        filename: object
+            String designing the relative path from where the dataset must be
+        loaded
+        nb_images: integer
+            Number of images that must be loaded (if None, the whole dataset is loaded)
+        """
+        with open(filename) as fp:
+            ds = json.load(fp)
+            self.image_size = ds["image_size"]
+            self.class_info = {int(k):ds["classes"][k] for k in ds["classes"].keys()}
```
garaud: There is a whitespace after the `:` when you create a dict.

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+            if nb_images is None:
+                self.image_info = {int(k):ds["images"][k] for k in ds["images"].keys()}
+            else:
+                self.image_info = {int(k):ds["images"][k] for k in ds["images"].keys() if int(k) < nb_images}
```
garaud: The `nb_images` threshold/filter works only if the keys of the images dataset can be converted to an integer, begin to 0 and are sorted. The `keys()` or dict iteration cannot return a sorted list of keys (it's not warranty by the dict implementation). For instance, `.keys()` could return `["10", "0", "3", "100"]`.

If we change the keys of the dataset, e.g. random letter and number, this won't work.

garaud: Maybe `enumerate` can do the job

```python
{k: v for idx, (k, v) in enumerate(ds["images"].items()) if idx < nb_images}
```

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+            for img_id, info in self.image_info.items():
+                info['labels'] = {int(k): v for k, v in info['labels'].items()}
+        utils.logger.info("The dataset has been loaded from {}".format(filename))
+
+class MapillaryDataset(Dataset):
+
+    def __init__(self, image_size, glossary_filename):
+        """ Class constructor
```
garaud: More explicit docstring for the class.

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        """
+        Dataset.__init__(self, image_size)
+        self.build_glossary(glossary_filename)
+
+    def build_glossary(self, config_filename):
+        """Read the Mapillary glossary stored as a json file at the data
+        repository root
+
+        Parameter:
```
garaud: About docstring : add a 's' and whitespace before `:`

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        ----------
+        config_filename: object
```
garaud: Write `str` instead of `object` for the parameter type description.

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+                       "classes": self.class_info,
+                       "images": self.image_info}, fp)
+        utils.logger.info("The dataset has been saved into {}".format(filename))
+
+    def load(self, filename, nb_images=None):
+        """Load a dataset from a json file indicated by `filename`
+
+        Parameter
+        ---------
+        filename: object
```
garaud: `str` instead of `object` in the parameter type description

sources/dataset.py
```diff
@@ -163,6 +139,73 @@ class (useful for semantic segmentation result printing)
...
+        Parameter:
+        ----------
+        config_filename: object
+            String designing the relative path of the dataset glossary
+        (based on Mapillary dataset)
+        """
+        with open(config_filename) as config_file:
+            glossary = json.load(config_file)
+        if "labels" not in glossary:
+            print("There is no 'label' key in the provided glossary.")
```
garaud: Maybe you want to use the logger instead of the print.

sources/dataset.py
```diff
@@ -261,52 +304,13 @@ def populate(self, output_dir, input_dir, nb_images=None, aggregate=False, label
...
-        utils.logger.info("The dataset has been loaded from {}".format(filename))
-
 class ShapeDataset(Dataset):
 
     def __init__(self, image_size, nb_classes):
         """ Class constructor
         """
-        self.image_size = image_size
-        self.class_info = defaultdict()
+        Dataset.__init__(self, image_size)
```
garaud: It's a better practice to use the `super` function when you write inheritance in Python.
Good to read: https://docs.python.org/3/library/functions.html#super

garaud (changes requested):

sources/dataset.py
delhomer: It guess `np.int` are involved. That was a kind of mystery for me: `cv2` method seems not compatible with such a format, hence the ugly casting.

sources/dataset.py
delhomer: You're write, it is in my implicit TODO list for a long time...

sources/dataset.py
garaud: Ok. Thanks

sources/dataset.py
delhomer: `image_info['labels']` is a dictionary such that each label id is the key of a 0/1 value which indicates whether or not there is the corresponding shapes on the picture.
For instance:
```
image_info['labels'] = {'0': 0, '1': 1, '2': 1}
```
In such a case, there are one object of type '1' (a circle) and one object of type '2' (a triangle), but no object of type '0' (a square).
As far as I understand your point, we can not use it as `image_info['labels']['type']`, as it may exist several shapes on the same image. It is similar with the Mapillary dataset, where several objects may exist on the same street-scene image. However, couldn't we consider that 0/1 integer behave just like boolean values?

sources/dataset.py
delhomer: Fixed with `super().__init__(image_size)`

sources/dataset.py
delhomer: Done.

sources/dataset.py
delhomer: Done.

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: Done

sources/dataset.py
delhomer: That's true, and the code is even uselessly complicated, as the following seems enough:
```
self.class_info = ds["classes"]
```

delhomer: Oh no correction: I wrote this complicated statements as the goal is to convert dictionary keys as integers (instead of strings). Anyway, I can drop the `.keys()`.

sources/dataset.py
delhomer: Defaultdicts have been turned into simple dicts.

sources/dataset.py
delhomer: Done.

sources/dataset.py
delhomer: Done.

sources/dataset.py
delhomer: Class attributes have been added to `ShapeDataset` definition as follows:
```
SQUARE = 0
CIRCLE = 1
TRIANGLE = 2
```
We can use them as in `image_info['labels'][self.SQUARE]` (for instance) if needed.

PR #30, opened 2018-03-28
Models

delhomer:
Integrate an example of implementation of [VGG](https://arxiv.org/abs/1409.1556) and [Inception](https://arxiv.org/abs/1409.4842) deep neural network architectures, for feature detection problem.


garaud (approved):
Looks good for me.

Issue #31, opened 2018-03-28
Make a setup.py


garaud:
Also find a name for the package. Some propositions:

- `from street.models import Segmentation`
- `from deeposlandia.models import Segmentation`
- `from streetscene.models import Segmentation`
- `from street_scene.models import Segmentation`

Note that the **shapes** dataset doesn't deal with street scene images. Is is relevant to name it "street*something"?

delhomer:
Ok to drop "street" from the package name. And from the project name as a consequence.
I like `deeposlandia`, we can imagine some other variants:
- `deeplandia`
- `oslanndia`
- `cnnlandia`
- `oslandcnn`
...

garaud:
I close this one.

PR #37, opened 2018-03-29
new setup.py file

garaud:
related to #31


garaud:
I think it's better if we merge all current pull requests and then, rename the `sources` folder and move some files, e.g. test files into the `tests` folder.

delhomer:
This looks great, thanks !

PR #39, opened 2018-03-30
Packet reorganization

delhomer:
Move the source code to `./deeposlandia/` in order to run the program as a standalone package. Move the test module to `./tests/`. Delete the `./sources/` repository.


garaud:
There are some `import utils` in the `train.py` and `test.py` files. I think we should replace them by a absolute or relative path

`from deeposlandia import utils`
or
`from .utils import stuff`

garaud (approved):
After the `import utils` update. Looks good for me.

PR #40, opened 2018-04-05
Keras generators

delhomer:
Integrate a new feature to Keras-based code: ability to build some data generators, either with Mapillary data or with random shapes. In latter case, some minor changes have been necessary, as the addition of a dummy class for representing background pixels.

A pair of tests has been introduced as well to consider the generator output shapes.


tests/test_generators.py
```diff
@@ -0,0 +1,57 @@
...
+    IMAGE_SIZE = 64
+    BATCH_SIZE = 10
+    datapath = ("data/shapes/preprocessed/" + str(IMAGE_SIZE)
+                + "_full/training")
+    config = utils.read_config(datapath + ".json")
+    gen = generator.create_generator(dataset, datapath, IMAGE_SIZE, BATCH_SIZE, config)
+    item = next(gen)
+    assert(len(item)==2)
+    im_shape = item[0].shape
+    assert(im_shape[0] == BATCH_SIZE)
```
garaud: You don't have to write parenthesis after the `assert`. Maybe you should write

`assert im_shapes[0] == BATCH_SIZE`

don't forget to write whitespaces before and after the `==`

Another comment: you can compare the shape instead of compare each element if you want, e.g.

`assert im_shape == (BATCH_SIZe, IMAGE_SIZE, IMAGE_SIZE, 3)`

tests/test_generators.py
```diff
@@ -0,0 +1,57 @@
...
+def test_mapillary_generator():
+    """Test the data generator for the Mapillary dataset
+
+    """
+    dataset = "mapillary"
+    IMAGE_SIZE = 256
+    BATCH_SIZE = 10
+    datapath = ("data/mapillary/preprocessed/" + str(IMAGE_SIZE)
+                + "_aggregated/training/")
+    config = utils.read_config("./data/mapillary/input/config_aggregate.json")
```
garaud: I think we should put a JSON file in the `tests` folder dedicated to the unit tests.

tests/test_generators.py
```diff
@@ -0,0 +1,57 @@
+"""Unit test associated to the generator building and feeding
+"""
+
+import numpy as np
```
garaud: Are these imports necessary?

tests/test_generators.py
```diff
@@ -0,0 +1,57 @@
+"""Unit test associated to the generator building and feeding
```
garaud: I prefer "related" than "associated".

Keep the name of the test file consistent with the name of the related module. I think you can remove the `s` at the end of the `test_generators.py` filename

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+        Image and labels Keras generators
+    """
+    generator = ImageDataGenerator()
+    im_generator = generator.flow_from_directory(
+        datapath,
+        classes=['images'],
+        target_size=(image_size, image_size),
+        batch_size=batch_size,
+        class_mode=None,
+        seed=SEED)
```
garaud: make the seed a parameter of the function, with a default value.

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+"""
+
+import numpy as np
+
+import keras as K
+from keras.preprocessing.image import ImageDataGenerator
+
+SEED = 1337
+
+def find_classes(img, config):
```
garaud: You also should test this function.

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Returns
+    -------
+    np.array
+        Output matrix
+    """
+    colors = [config['classes'][sc]['color'] for sc in config['classes']]
+    nb_classes = len(config['classes'])
+    mask = np.zeros(img.shape[:3]) + nb_classes
+    for class_id in range(nb_classes):
```
garaud: It's not really the id of each class? I suggest `idx` or `class_idx`

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+        for i in range(img.shape[0]):
+            mask[i][[[img[i, x, y, ::-1].tolist() == colors[class_id] for y in range(img.shape[2])]
+                   for x in range(img.shape[1])]] = class_id
+    return mask
```
garaud: For my point of a view, a "mask" array is an array with booleans. You can use it with something like `a[mask]`.

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+    ----------
+    img : np.array
+        Input RGB images
+    config : dict
+        Dataset glossary
+
+    Returns
+    -------
+    np.array
+        Output matrix
```
garaud: Output matrix -> array with the index of each class for each pixel shape (batch, size, size, num_classes)

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+    if dataset == 'shapes':
+        mask_generator = generator.flow_from_directory(
+            datapath,
+            classes=['labels'],
+            target_size=(image_size, image_size),
+            batch_size=batch_size,
+            class_mode=None,
+            seed=SEED)
+    else:
+        mask_generator = generator.flow_from_directory(
```
garaud: Maybe I would have been changed the name of these variable from my previous copy/pasting for this snippet. `labels_generator` is clearer than `masks_generator`.

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+    image_size : integer
+        Number of width (resp. height) pixels
+    batch_size : integer
+        Number of images in each training batch
+    config : dict
+        Dataset glossary
+
+    Returns
+    -------
+    zip
```
garaud: `zip` it's not a type. This is a function name which returns a generator. Maybe "generator" or "iterable". We can be more specific with a short sentence such as "iterate over the input image batch and label bach"...

deeposlandia/generator.py
```diff
@@ -0,0 +1,119 @@
...
+    -------
+    np.array
+        Output matrix
+    """
+    colors = [config['classes'][sc]['color'] for sc in config['classes']]
+    nb_classes = len(config['classes'])
+    mask = np.zeros(img.shape[:3]) + nb_classes
+    for class_id in range(nb_classes):
+        for i in range(img.shape[0]):
+            mask[i][[[img[i, x, y, ::-1].tolist() == colors[class_id] for y in range(img.shape[2])]
```
garaud: I don't get the `::-1`. You have to inverse the order of some elements?

```python
a = np.arange(24).reshape((4, 6))
a[1, ::-1]
# array([11, 10,  9,  8,  7,  6])
a[1, :]
# array([ 6,  7,  8,  9, 10, 11])
```

garaud (changes requested):
Some comments. Maybe I would add some comments about `find_classes`.

tests/test_generators.py
delhomer: Ok, modified.

tests/test_generators.py
delhomer: There are not, I remove them in both test modules.

tests/test_generators.py
delhomer: Done.

deeposlandia/generator.py
delhomer: Yes... It depends on the way we read the image. They are stored with OpenCV in a (B,G,R) format. However most draft tests have consisted in reading images with Pillow, which is simpler. And Pillow considers the (R, G, B) format...

However, I am not able to say which format is chosen by Keras `flow_frow_directory` method. This point needs to be vigilant regarding the input data we get.

deeposlandia/generator.py
delhomer: Done.

deeposlandia/generator.py
delhomer: Ok, got it, I change the variable name.

deeposlandia/generator.py
delhomer: Ok for the description updating.

deeposlandia/generator.py
delhomer: The name is now `label_generator`.

deeposlandia/generator.py
delhomer: I've added a comment about this point in the docstring.

deeposlandia/generator.py
delhomer: The docstring has been modified.

deeposlandia/generator.py
delhomer: I am not sure to really understand this point. I have standardized the way of iterating over the label ids (just did as in `to_categorical` method : `for idx, class_id in enumerate(label_ids):`). However the fact that the good labels are put in the good array indices is not clear to me.

tests/test_generators.py
delhomer: We should put some image files as well, don't we ?

garaud:
Yes, you're right.
Raphael Delhome writes:

>
>
>
>> +
> +import numpy as np
> +import keras as K
> +
> +from deeposlandia import generator, network, utils
> +
> +def test_mapillary_generator():
> +    """Test the data generator for the Mapillary dataset
> +
> +    """
> +    dataset = "mapillary"
> +    IMAGE_SIZE = 256
> +    BATCH_SIZE = 10
> +    datapath = ("data/mapillary/preprocessed/" + str(IMAGE_SIZE)
> +                + "_aggregated/training/")
> +    config = utils.read_config("./data/mapillary/input/config_aggregate.json")
>
> We should put some image files as well, don't we ?


-- 
Damien Garaud
http://oslandia.com

delhomer:
Some modifications have been implemented since your last comments:
- Labels of Shapes dataset are now stored as one-channelled images, with values corresponding to their class ids.
- Consequently we do not need the `find_classes` method any more, it has been dropped.
- The distinction between both dataset is done in `create_generator` method, this distinction is only done on `label_ids` recovering.
- Consequently changes to `to_categorical` method have been cancelled, the method is just as you provided it in `segnet_keras.py`.
- A test dedicated to the `to_categorical` method has been introduced.
- Some data has been added into the `tests/data` repository, in order to make the tests replicable.

garaud (approved):

garaud:
Looks good for me.

PR #41, opened 2018-04-06
Keras model building

delhomer:
Implement dedicated classes to build feature detection and semantic segmentation architectures. It is crucial to notice that the produced classes correspond to **architectures** instead of **models** for now. We will use the Keras API to easily create models, by considering `Network.X` and `Network.Y` : 
```
net = FeatureDetectionNetworl(image_size=224, nb_channels=3, nb_labels=65)
model = keras.models.Model(net.X, net.Y)
model.compile(...)
model.summary() # Gives the whole architecture of the network, built during the instanciation of `net`
```

In this PR, a first example of state-of-the-art model architecture reusing is also provided with `VGG16` network (see *Simonyan, 2015*). For more details, see `architecture` attribute of class `FeatureDetectionNetwork` and `SemanticSegmentation`.

Some unit tests come with these new features.

### DIFF
--- a/deeposlandia/keras_feature_detection.py
+++ b/deeposlandia/keras_feature_detection.py
@@ -1,0 +1,113 @@
+
+"""Design a feature detection model with Keras API
+"""
+
+import keras as K
+
+from deeposlandia.network import ConvolutionalNeuralNetwork
+
+class FeatureDetectionNetwork(ConvolutionalNeuralNetwork):
+    """
+    """
+
+    def __init__(self, network_name="mapillary", image_size=512, nb_channels=3,
+                 nb_labels=65, learning_rate=1e-4, architecture="simple"):
+        ConvolutionalNeuralNetwork.__init__(self, network_name, image_size,
+                                            nb_channels, nb_labels, learning_rate)
+        if architecture == "vgg16":
+            self.Y = self.vgg16()
+        elif architecture == "vgg19":
+            self.Y = self.vgg19()
+        elif architecture == "inception-v1":
+            self.Y = self.inception_v1()
+        elif architecture == "inception-v2":
+            self.Y = self.inception_v2()
+        elif architecture == "inception-v3":
+            self.Y = self.inception_v3()
+        elif architecture == "inception-v4":
+            self.Y = self.inception_v4()
+        elif architecture == "Xception":
+            self.Y = self.xception()
+        elif architecture == "resnet":
+            self.Y = self.resnet()
+        else:
+            self.Y = self.simple()
+
+    def output_layer(self, x, depth):
+        """Build an output layer to a neural network, as a dense layer with sigmoid activation
+        function as the point is to detect multiple labels on a single image
+
+        Parameters
+        ----------
+        x : tensor
+            Previous layer within the neural network (last hidden layer)
+        depth : integer
+            Dimension of the previous neural network layer
+
+        Returns
+        -------
+        tensor
+            2D output layer
+        """
+        y = K.layers.Dense(depth, name='output_fc')(x)
+        y = K.layers.Activation(activation='sigmoid', name='output_activation')(y)
+        return y
+
+    def simple(self):
+        """Build a simple default convolutional neural network composed of 3 convolution-maxpool
+        blocks and 1 fully-connected layer
+
+        Returns
+        -------
+        tensor
+            Output predictions, that have to be compared with ground-truth values
+        """
+        layer = self.convolution(self.X, nb_filters=16, kernel_size=7, name='conv1')
+        layer = self.maxpool(layer, pool_size=2, strides=2, name='pool1')
+        layer = self.convolution(layer, nb_filters=32, kernel_size=5, name='conv2')
+        layer = self.maxpool(layer, pool_size=2, strides=2, name='pool2')
+        layer = self.convolution(layer, nb_filters=64, kernel_size=3, name='conv3')
+        layer = self.maxpool(layer, pool_size=2, strides=2, name='pool3')
+        layer = self.flatten(layer, name='flatten1')
+        layer = self.dense(layer, depth=512, dropout_rate=0.75, name='fc1')
+        return self.output_layer(layer, depth=self._nb_labels)
+
+    def vgg16(self):
+        """
+        """
+        pass
+
+    def vgg19(self):
+        """
+        """
+        pass
+
+    def inception_v1(self):
+        """
+        """
+        pass
+    
+    def inception_v2(self):
+        """
+        """
+        pass
+
+    def inception_v3(self):
+        """
+        """
+        pass
+
+    def inception_v4(self):
+        """
+        """
+        pass
+
+    def xception(self):
+        """
+        """
+        pass
+
+    def resnet(self):
+        """
+        """
+        pass
--- a/deeposlandia/keras_feature_detection.py
+++ b/deeposlandia/keras_feature_detection.py
@@ -2,8 +2,10 @@
 """Design a feature detection model with Keras API
 """
 
-import keras as K
+import h5py
 
+import keras as K
+from keras.applications import VGG16
 from deeposlandia.network import ConvolutionalNeuralNetwork
 
 class FeatureDetectionNetwork(ConvolutionalNeuralNetwork):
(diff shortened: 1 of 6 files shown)


tests/test_feature_detection.py
```diff
@@ -0,0 +1,41 @@
+"""Unit test related to feature detection model instanciation
```
garaud: I would have written "Unit test related to the feature detection model"

I don't think that the word `instanciation` is correct.

tests/test_feature_detection.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+from keras.models import Model
+
+from deeposlandia.keras_feature_detection import FeatureDetectionNetwork
+
+def test_simple_network_architecture():
+    """Test the instanciation of a simple feature detection network
```
garaud: Could you remove the word `instanciation`?

tests/test_feature_detection.py
```diff
@@ -0,0 +1,41 @@
...
+    m = Model(net.X, net.Y)
+    input_shape = m.input_shape
+    output_shape = m.output_shape
+    assert len(input_shape) == 4
+    assert input_shape[1:] == (IMAGE_SIZE, IMAGE_SIZE, NB_CHANNELS)
+    assert len(output_shape) == 2
+    assert output_shape[1] == NB_LABELS
+
+def test_vgg16_network_architecture():
+    """Test the instanciation of a VGG16-inspired feature detection network
```
garaud: Could you remove the word "instanciation"?

tests/test_semantic_segmentation.py
```diff
@@ -0,0 +1,23 @@
+"""Unit test related to semantic segmentation model instanciation
```
garaud: Unit test related to the semantic segmentation model.

tests/test_semantic_segmentation.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from keras.models import Model
+
+from deeposlandia.keras_semantic_segmentation import SemanticSegmentationNetwork
+
+def test_network_instanciation():
+    """Test the instanciation of a simple feature detection network
```
garaud: Could you remove the word "instanciation"?

tests/test_feature_detection.py
```diff
@@ -0,0 +1,41 @@
...
+                                  nb_channels=NB_CHANNELS, nb_labels=NB_LABELS)
+    m = Model(net.X, net.Y)
+    input_shape = m.input_shape
+    output_shape = m.output_shape
+    assert len(input_shape) == 4
+    assert input_shape[1:] == (IMAGE_SIZE, IMAGE_SIZE, NB_CHANNELS)
+    assert len(output_shape) == 2
+    assert output_shape[1] == NB_LABELS
+
+def test_vgg16_network_architecture():
```
garaud, line 25: It seems that this test is quite long (several seconds). What do you think? Do you know why?

deeposlandia/network.py
```diff
@@ -5,20 +5,32 @@
 
 class ConvolutionalNeuralNetwork:
     """Convolutional neural network design
+
+    Attributes
+    ----------
+    _network_name : str
```
garaud: It's quite strange for me to have some "protected" attributes in the docstring. The most of the time, you write `self._attr` when you don't want to let it be "public", so you don't write a docstring for it.

I prefer to have "public" and explicit attributes which define you typed object. Is it possible to replace the `self._attr` by `self.attr`?

deeposlandia/keras_feature_detection.py
```diff
@@ -0,0 +1,96 @@
...
+        Number of classes in the dataset glossary
+    X : tensor
+        (batch_size, image_size, image_size, nb_channels)-shaped input tensor; input image data
+    Y : tensor
+        (None, nb_classes)-shaped output tensor, built as the output of the last network layer
+    """
+
+    def __init__(self, network_name="mapillary", image_size=512, nb_channels=3,
+                 nb_labels=65, learning_rate=1e-4, architecture="simple"):
+        ConvolutionalNeuralNetwork.__init__(self, network_name, image_size,
```
garaud: I think you should use the `super()` function for the inheritance.

deeposlandia/network.py
```diff
@@ -5,20 +5,32 @@
...
         self._network_name = network_name
         self._image_size = image_size
         self._nb_channels = nb_channels
         self._nb_labels = nb_labels
-        self._X = K.layers.Input(shape=(image_size, image_size, nb_channels))
+        self.X = K.layers.Input(shape=(image_size, image_size, nb_channels), name="input")
 
     def convolution(self, x, nb_filters, kernel_size, strides=1, padding="same",
-                    activation='relu', batch_norm=True):
+                    activation='relu', batch_norm=True, name=""):
```
garaud: The Keras layer, by default, defines a layer name according to its type, e.g. `Conv2D`, with an incremented integer such as `conv2d_2`, `conv2d_3`, and so on.

I think we should set it `None` by default. Then, Keras defines its own. And if we want to specify it, we can.

deeposlandia/network.py
```diff
@@ -137,17 +155,35 @@ def dense(self, x, depth, dropout_rate=1.0, activation='relu', batch_norm=True):
...
         if batch_norm:
-            x = K.layers.BatchNormalization()(x)
-        x = K.layers.Activation(activation)(x)
-        x = K.layers.Dropout(dropout_rate)(x)
+            x = K.layers.BatchNormalization(name=name+'_bn')(x)
+        x = K.layers.Activation(activation, name=name+'_activation')(x)
+        x = K.layers.Dropout(dropout_rate, name=name+'_dropout')(x)
         return x
 
+    def flatten(self, x, name=""):
```
garaud: Same comment about the `name` parameter.

garaud (changes requested):
The `Model(nn.X, nn.Y)` is a very good idea. Some comment with minor changes.

deeposlandia/network.py
delhomer: Yes, these stuffs are remainders of a quite clumsy C++-convention adaptation. Useless in Python, I remove them.

deeposlandia/network.py
delhomer: Yes indeed, I saw the Keras default behavior for layer naming. `None` is a better idea than empty string, for sure. I change this.

deeposlandia/network.py
delhomer: Seen.

tests/test_feature_detection.py
delhomer: `instanciation` word has been removed.

tests/test_feature_detection.py
delhomer: Done.

tests/test_feature_detection.py
delhomer: Done.

tests/test_semantic_segmentation.py
delhomer: Done.

tests/test_semantic_segmentation.py
delhomer: Done.

deeposlandia/keras_feature_detection.py
delhomer: Changed to `super().__init__(network_name, image_size, ...)`.
This format is only `Python-3.x` compliant, however I guess the code is still Python3-oriented. We mention Python3 in the `setup.py`, I do not if we have to specify again elsewhere.

tests/test_feature_detection.py
delhomer: Hum, I'm quite surprised, it is less than one second on my dev machine.

However, the fact it can be long is not surprising: we are instantiating a deep neural network with around 130 millions parameters here. Some time is needed to create and initialize all the weights, I guess.

garaud:
We mention Python 3 in the README.md file
Raphael Delhome writes:

>
>
>
>> +        Name of the network
> +    _image_size : integer
> +        Input image size (height and width are equal)
> +    _nb_channels : integer
> +        Number of input image channels (1 for greyscaled images, 3 for RGB images)
> +    _nb_labels : integer
> +        Number of classes in the dataset glossary
> +    X : tensor
> +        (batch_size, image_size, image_size, nb_channels)-shaped input tensor; input image data
> +    Y : tensor
> +        (None, nb_classes)-shaped output tensor, built as the output of the last network layer
> +    """
> +
> +    def __init__(self, network_name="mapillary", image_size=512, nb_channels=3,
> +                 nb_labels=65, learning_rate=1e-4, architecture="simple"):
> +        ConvolutionalNeuralNetwork.__init__(self, network_name, image_size,
>
> Changed to `super().__init__(network_name, image_size, ...)`.
> This format is only `Python-3.x` compliant, however I guess the code is still Python3-oriented. We mention Python3 in the `setup.py`, I do not if we have to specify again elsewhere.


-- 
Damien Garaud
http://oslandia.com

delhomer:
I pushed another commit so as to fix the case when layer names are equal to `None`.

Now, if a hand-made layer has a `block_name` equal to None, all the layers that are contained into it take `None` as name (*e.g.* for `convolution()`, that contains a `Conv2D`, a `BatchNormalization()` and a `Activation()` native layers).

PR #47, opened 2018-04-11
improve uni tests for the  generators module

### LINKED ISSUE #44: Avoid some warning messages about elementwise operations
When I launch the test with `pytest`, I've got these messages:

```
=================== warnings summary =========================
tests/test_generators.py::test_featdet_shape_generator
  deeposlandia/deeposlandia/generator.py:24: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
    mask = img == label

tests/test_generators.py::test_semseg_shape_generator
  deeposlandia/deeposlandia/generator.py:49: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
    mask = img == label

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

garaud:
* check the type and value of each label ids
* add a function to carry out this checking
* add a unit test

also fix #44

### DIFF
--- a/deeposlandia/generator.py
+++ b/deeposlandia/generator.py
@@ -6,6 +6,22 @@ import numpy as np
 from keras.preprocessing.image import ImageDataGenerator
 
 
+def check_label_id(labels, label_ids):
+    """Check the labels_ids
+
+    Parameters
+    ----------
+    labels : numpy.array
+        Array of labels ids (contain in a image for instance)
+    label_ids : list of integers
+        List of label ids
+    """
+    labels = np.unique(labels)
+    for label in labels:
+        assert label in label_ids, "array values must be ids specified in label_ids"
+    return True
+
+
 def feature_detection_labelling(img, label_ids):
     """One-hot encoding for feature detection problem
 
@@ -20,6 +36,7 @@ def feature_detection_labelling(img, label_ids):
     numpy.array
         Label encoding, array of shape (batch_size, nb_labels)
     """
+    check_label_id(img, label_ids)
     for idx, label in enumerate(label_ids):
         mask = img == label
         img[mask] = idx
--- a/deeposlandia/generator.py
+++ b/deeposlandia/generator.py
@@ -146,9 +146,9 @@ def create_generator(dataset, model, datapath, image_size, batch_size, config, s
     image_generator = feed_generator(datapath, "images", image_size, batch_size, seed)
     label_generator = feed_generator(datapath, "labels", image_size, batch_size, seed)
     if model == 'feature_detection':
-        label_generator = iter(feature_detection_labelling(x, label_ids)
-                               for x in label_generator)
+        label_generator = (feature_detection_labelling(x, label_ids)
+                           for x in label_generator)
     else:
-        label_generator = iter(semantic_segmentation_labelling(x, label_ids)
-                               for x in label_generator)
+        label_generator = (semantic_segmentation_labelling(x, label_ids)
+                           for x in label_generator)
     return zip(image_generator, label_generator)
--- a/deeposlandia/generator.py
+++ b/deeposlandia/generator.py
@@ -3,8 +3,8 @@
 
 import numpy as np
 
-import keras as K
 from keras.preprocessing.image import ImageDataGenerator
+
 
 def feature_detection_labelling(img, label_ids):
     """One-hot encoding for feature detection problem
@@ -26,6 +26,7 @@ def feature_detection_labelling(img, label_ids):
     flattened_images = img.reshape(img.shape[0], -1).astype(np.uint8)
     one_hot_encoding = np.eye(len(label_ids))[flattened_images]
     return one_hot_encoding.any(axis=1)
+
 
 def semantic_segmentation_labelling(img, label_ids):
     """One-hot encoder for semantic segmentation problem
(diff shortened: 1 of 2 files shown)


deeposlandia/generator.py
```diff
@@ -3,9 +3,26 @@
...
+    Parameters
+    ----------
+    labels : numpy.array
+        Array of labels ids (contain in a image for instance)
+    label_ids : list of integers
+        List of label ids
+    """
+    labels = np.unique(labels)
+    for label in labels:
+        if label not in label_ids:
```
delhomer: May we write the following line, instead of lines 20 and 21?
```
assert label in label_ids, "Array values must be IDs specified in label_ids"
```

deeposlandia/generator.py
```diff
@@ -117,17 +139,17 @@ def create_generator(dataset, model, datapath, image_size, batch_size, config, s
 
     """
     if dataset == 'shapes':
-        label_ids = [c for c in config['classes']]
+        label_ids = [int(c) for c in config['classes']]
```
delhomer, line 141: Do we keep this casting operation if the dataset definition is updated? You would say, it is better to keep it as a safety net... (I agree.)

deeposlandia/generator.py
```diff
@@ -117,17 +139,17 @@ def create_generator(dataset, model, datapath, image_size, batch_size, config, s
...
     if model == 'feature_detection':
-        label_generator = iter(feature_detection_labelling(x, label_ids)
-                               for x in label_generator)
+        label_generator = (feature_detection_labelling(x, label_ids)
+                           for x in label_generator)
     else:
-        label_generator = iter(semantic_segmentation_labelling(x, label_ids)
-                               for x in label_generator)
+        label_generator = (semantic_segmentation_labelling(x, label_ids)
+                           for x in label_generator)
```
delhomer: Hum, I just didn't know we could drop the `iter` part. Nice!

tests/test_generators.py
```diff
@@ -20,9 +22,35 @@ def test_feature_detection_labelling():
     b = generator.feature_detection_labelling(a, range(MIN, MAX))
     assert b.shape == (BATCH_SIZE, MAX)
 
+
+def test_feature_detection_labelling_wrong_label_id():
```
delhomer, line 26: I am wondering if such a test function could be implemented for semantic segmentation as well?

tests/test_generators.py
```diff
@@ -20,9 +22,35 @@ def test_feature_detection_labelling():
...
+                          [10, 10, 0, 0, 0],
+                          [10, 10, 0, 0, 0]])
+    labels = np.array(one_label.tolist() + two_label.tolist())
+    labels = labels.reshape((2, 4, 5, 1))
+    # do not allow a list of string as label ids
+    with pytest.raises(AssertionError):
+        b = generator.feature_detection_labelling(labels, ['0', '3', '10'])
+
+    # label_id=2 is a wrong id. All values=3 are kept and used as index value. But
+    # there are only 3 classes, so a list of index [0, 1, 2].
```
delhomer: I do not understand the second and third sentences in this comment. Could it be possible to make it clearer?

delhomer (changes requested):
Almost fine for me. In my opinion, a testing function for semantic segmentation labelling values is still missing. Apart from that, well done!

deeposlandia/generator.py
garaud: good point.

deeposlandia/generator.py
garaud: Nope. But, for now, this makes it work.

tests/test_generators.py
garaud: done.

tests/test_generators.py
garaud: done.

tests/test_generators.py
```diff
@@ -108,3 +152,24 @@ def test_semseg_shape_generator():
...
+    model = "conquer_the_world"
+    IMAGE_SIZE = 10
+    BATCH_SIZE = 10
+    datapath = ("./tests/data/" + dataset + "/training")
+    config = {"classes": list('0123')}
+
+    # wrong dataset name
+    with pytest.raises(ValueError) as excinfo:
+        generator.create_generator(dataset, model, datapath, IMAGE_SIZE, BATCH_SIZE, config)
+    assert str(excinfo.value) == "Wrong dataset name fake"
```
delhomer:
```
assert str(excinfo.value) == "Wrong dataset name: {}".format(dataset)
```
(or any other valid string format in Python 3.5)

tests/test_generators.py
```diff
@@ -108,3 +152,24 @@ def test_semseg_shape_generator():
...
+
+    # wrong model name
+    with pytest.raises(ValueError) as excinfo:
+        generator.create_generator('shapes', model, datapath, IMAGE_SIZE, BATCH_SIZE, config)
+    assert str(excinfo.value) == "Wrong model name conquer_the_world"
```
delhomer:
```
assert str(excinfo.value) == "Wrong model name: {}".format(model)
```
(or any other valid string format in Python 3.5)

delhomer (changes requested):
Two minor comments in new test functions.

After that, It's fine for merging!

tests/test_generators.py
garaud: done

tests/test_generators.py
garaud: done

garaud:
I also modify the string formatting

```python
f"Wrong model name {model}"
```

by 

```python
"Wrong model name {}".format(name)
```

The f-string format is only for python 3.6

PR #48, opened 2018-04-11
Data preprocessing

delhomer:
This PR update the Dataset definition, in the sense that labels are now called `labels` instead of `classes`, for a sake of clarity. In addition to this detail, the labels are now stored as a list of dictionaries (one element per labels), instead of a dictionary of dictionaries where the label IDs played the role of dictionary keys. So do the image representation.

This behavior actually fits those of Mapillary configuration file `config.json`, as well as those of `config_aggregate.json`.

Furthermore, sample data used in tests (stored in `tests/data/`) has been updated according to these changes.

From now, there is no reason to distinguish between datasets when creating data generators (some additional commits will certainly be necessary in addition to the PR #47 , not yet merged at this time).

Preprocessed versions of images should be generated again with these new modifications. :warning: In the case of aggregated datasets, the label IDs are not original IDs any more, but they are reindexed from `0` to `N-1`, with `N` being the number of aggregated labels.


garaud:
Looks good for me.

PR #49, opened 2018-04-16
Datagen

delhomer:
This PR introduces a new module dedicated to the dataset creation. Before to train a model on a given dataset, we choose to write the dataset on the file system, for further reusing. The following command is then expected, as an example:
```
python deeposlandia/datagen.py -d mapillary -s 224 -a -t 200 -v 20 -T 50
```
In this example, the program will generate 200 training images, 20 validation images and 50 testing images for the Mapillary dataset, and consequently produce 224*224 preprocessed images. As a remark, testing dataset generation implies only image generation, without labels.

Some additional tests has been developed as well to ensure the effectiveness of dataset creation, populating and loading operations (see `tests/test_dataset.py`). Concurrently a new data sample have been added to the `tests` repo, so as to accomplish the new tests.

From now, we do consider that dataset creation is not in the scope of the training program. Consequently, instead of creating the dataset before the training process if the former does not exist on the file system, the latter will just crash by telling there is no dataset (and eventually, by advising to launch the corresponding dataset creation command).


deeposlandia/datagen.py
```diff
@@ -0,0 +1,118 @@
...
+                        help=("Number of training images"))
+    parser.add_argument('-v', '--nb-validation-image',
+                        type=int,
+                        default=200,
+                        help=("Number of validation images"))
+    return parser
+
+
+if __name__=='__main__':
+    """Main method: generate a dataset from some instance specifications
```
garaud: I think the string here is not necessary. This is not a docstring. If we want to keep it, we can turn this content into comments with `#`.

deeposlandia/datagen.py
```diff
@@ -0,0 +1,118 @@
+"""Main method to generate new datasets
+"""
```
garaud: Maybe you can add an usage example of this script, e.g.

```to generate data about the xxx dataset, do

python datagen.py --args ...
```

tests/test_dataset.py
```diff
@@ -0,0 +1,111 @@
+"""Unit test related to the dataset creation, population and loading
+"""
+
+import json
+import numpy as np
```
garaud: numpy is imported but not used.

tests/test_dataset.py
```diff
@@ -0,0 +1,111 @@
...
+    config_filename = "tests/data/mapillary/config_aggregate.json"
+    with open(config_filename) as fobj:
+        config = json.load(fobj)
+    NB_LABELS = len(config["labels"])
+    d = MapillaryDataset(IMAGE_SIZE, config_filename)
+    assert d.image_size == IMAGE_SIZE
+    assert d.get_nb_labels(see_all=True) == NB_LABELS
+    assert d.get_nb_images() == 0
+
+def test_mapillary_dataset_population():
```
garaud: When you test the creation of several image files from a specific dataset, you modify the working directory. See the `git status` after a `pytest tests/test_dataset.py`.

I think, we should have:

* a directory to input data (JSON, some images, etc.)
* a directory to output data which has to be cleaned after each unit test. I think we can add some pytest fixtures to create output dir on the fly, generate files, check some stuff, and remove the output dir.

garaud (changes requested):

deeposlandia/datagen.py
delhomer: The line has been dropped.

deeposlandia/datagen.py
delhomer: Two examples have been added to the module docstring.

tests/test_dataset.py
delhomer: Import removed.

tests/test_dataset.py
delhomer: Ok, I guess that for now, the output of the corresponding tests has been committed. You think it could be better to create the image files and delete it on the fly, without committing them?

garaud:
I think we cannot have commited test files such as JSON and images which
will be modified every time we launch the test.

And I think it's better to create and clean-up for each test.
Raphael Delhome writes:

>
>
>
>> +    assert d.get_nb_labels() == 1
> +
> +def test_mapillary_dataset_creation():
> +    """Create a Mapillary dataset from a configuration file
> +    """
> +    IMAGE_SIZE = 224
> +    config_filename = "tests/data/mapillary/config_aggregate.json"
> +    with open(config_filename) as fobj:
> +        config = json.load(fobj)
> +    NB_LABELS = len(config["labels"])
> +    d = MapillaryDataset(IMAGE_SIZE, config_filename)
> +    assert d.image_size == IMAGE_SIZE
> +    assert d.get_nb_labels(see_all=True) == NB_LABELS
> +    assert d.get_nb_images() == 0
> +
> +def test_mapillary_dataset_population():
>
> Ok, I guess that for now, the output of the corresponding tests has been committed. You think it could be better to create the image files and delete it on the fly, without committing them?


-- 
Damien Garaud
http://oslandia.com

delhomer:
Some changes has been brought with last commits:
- a training data sample is kept on the git repo, it is used for a wide range of tests (regarding datasets, generators and even model training).
- Mapillary raw data sample for dataset population testing is kept as well in a dedicated repository (`tests/data/mapillary/input_dataset_population` instead of `tests/data/mapillary/validation/raw/`)
- images that were created during dataset population tests are stored into `tests/data/dataset/tmp/` during the test, and deleted on-the-fly at the end of the tests

We could have replicated the data file tree (see issue #23 ), however as you said, files that are generated during tests are temporary per nature, most of the time. I hope it is OK with these changes.

deeposlandia/datagen.py
```diff
@@ -0,0 +1,129 @@
+"""Main method to generate new datasets
+
+Example of program calls:
+
+- generate 64*64 pixel images from Shapes dataset, 10000 images in the training set, 100 in the
+validation set, 1000 in the testing set:
+```
```
garaud: the docstring python syntax is often written in ReST, [Restructured
Text](http://docutils.sourceforge.net/rst.html) instead of markdow.

See http://docutils.sourceforge.net/docs/user/rst/quickstart.html

The ReST format
to write snippet of code can be:

```
The following snippet::

    import sys
    sys.prefix
```

with two `:` and at least four indentation whitespaces.

And if you want to inline it, you have to use two ticks such as:

```
you can call the function ``foobar``.
```

tests/test_dataset.py
garaud: I previously mentioned the [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html). I think it's a proper and cleaner way to handle temporary directories. See the following example https://docs.pytest.org/en/latest/tmpdir.html

garaud (changes requested):

deeposlandia/datagen.py
delhomer: Ok, modified.

garaud:
despite the conflict with the module `generator.py` as specified by Github, it's good for me.

garaud (approved):

PR #50, opened 2018-04-17
Model training

delhomer:
This PR introduces the main refactoring of legacy Tensorflow-based code: it provides an executable Python module to undertake some neural network model training and inference.

The API has been slightly modified, please consider for instance this simple command, that will train a feature detection model for 3 epochs on Shapes dataset (64*64-pixel images):
```
python deeposlandia/kerastrain.py -M feature_detection -D shapes -s 64 -e 3
```
The number of images can be directly modified as well, see *e.g.* a very-fast dummy training with 1000 training images, 10 validation images and 100 test images:
```
python deeposlandia/kerastrain.py -M feature_detection -D shapes -s 64 -e 3 -it 1000 -iv 10 -ii 100
```
In such a case, dataset generator are fed with the whole dataset, however the training iteration stops when the corresponding amount of images has been scanned.

Basic tests have been developed as well to go with this evolution.

### DIFF
--- a/deeposlandia/kerastrain.py
+++ b/deeposlandia/kerastrain.py
@@ -1,0 +1,243 @@
+"""Main method to train neural networks with Keras API
+"""
+
+import argparse
+import os
+import sys
+import numpy as np
+
+from keras import backend
+from keras.models import Model
+
+from deeposlandia import generator, utils
+from deeposlandia.keras_feature_detection import FeatureDetectionNetwork
+from deeposlandia.keras_semantic_segmentation import SemanticSegmentationNetwork
+
+def add_instance_arguments(parser):
+    """Add instance-specific arguments from the command line
+
+    Parameters
+    ----------
+    parser : argparse.ArgumentParser
+        Input parser
+
+    Returns
+    -------
+    argparse.ArgumentParser
+        Modified parser, with additional arguments
+    """
+    parser.add_argument('-a', '--aggregate-label', action='store_true',
+                        help="Aggregate labels with respect to their categories")
+    parser.add_argument('-D', '--dataset',
+                        required=True,
+                        help="Dataset type (either mapillary or shapes)")
+    parser.add_argument('-M', '--model',
+                        default="feature_detection",
+                        help=("Type of model to train, either "
+                              "'feature_detection' or 'semantic_segmentation'"))
+    parser.add_argument('-n', '--name',
+                        default="cnn",
+                        help=("Model name that will be used for results, "
+                              "checkout and graph storage on file system"))
+    parser.add_argument('-N', '--network',
+                        default='simple',
+                        help=("Neural network size, either 'simple', 'vgg' or "
+                              "'inception' ('simple' refers to 3 conv/pool "
+                              "blocks and 1 fully-connected layer; the others "
+                              "refer to state-of-the-art networks)"))
+    parser.add_argument('-p', '--datapath',
+                        default="./data",
+                        help="Relative path towards data directory")
+    return parser
+
+def add_hyperparameters(parser):
+    """Add hyperparameter arguments from the command line
+
+    Parameters
+    ----------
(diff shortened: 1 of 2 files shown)


tests/test_model_training.py
```diff
@@ -0,0 +1,50 @@
...
+    assert len(hist.history) == 2
+    assert all(k in hist.history.keys() for k in ['acc', 'loss'])
+    assert hist.history['acc'][0] >= 0 and hist.history['acc'][0] <= 1
+
+    test_image = np.random.randint(0, 255, [BATCH_SIZE, IMAGE_SIZE, IMAGE_SIZE, 3])
+    score = model.predict(test_image)
+    assert score.shape == (BATCH_SIZE, len(label_ids))
+    assert all(0 <= s and s <= 1 for s in score.ravel())
+
+    BACKUP_PATH = "./tests/data/" + dataset + "/backups/"
```
garaud: you can use the `path` function from the `os.path` module to concatenate paths.

deeposlandia/kerastrain.py
```diff
@@ -0,0 +1,226 @@
...
+    return parser
+
+if __name__=='__main__':
+    """Main method: run a convolutional neural network using Keras API
+    """
+
+    # Parse command-line arguments
+    parser = argparse.ArgumentParser(description=("Convolutional Neural Netw"
+                                                  "ork on street-scene images"))
+    parser = add_instance_arguments(parser)
```
garaud, line 127: split the arg parser into different functions is a good idea! Well done.

deeposlandia/kerastrain.py
```diff
@@ -0,0 +1,226 @@
...
+                                    args.image_size, args.model)
+
+    # Instance name (name + image size + network size + batch_size
+    # + aggregate? + dropout + learning_rate)
+    instance_args = [args.name, args.image_size, args.network, args.batch_size,
+                     aggregate_value, args.dropout, utils.list_to_str(args.learning_rate)]
+    instance_name = utils.list_to_str(instance_args, "_")
+
+    # Data gathering
+    if (os.path.isfile(folders["training_config"]) and os.path.isfile(folders["validation_config"])
```
garaud, line 144: Does it imply that if you don't have (or don't want to have) a testing dataset, this won't work?

garaud (changes requested):
Same comment about generated files for testing purposes: create and clean-up for each test_* functions.

deeposlandia/kerastrain.py
delhomer: Thanks. :)
Actually that was a real mess, we need to be more organized. I hope it is good enough!

deeposlandia/kerastrain.py
delhomer: Thanks. :)
Actually that was a real mess, we need to be more organized. I hope it is good enough!

deeposlandia/kerastrain.py
delhomer: Good question. At this point, we need to have a testing dataset, that is true. I considered that we generally have such a dataset, either by gathering third-part data, or by generating our own dataset.

The point is that we need to have a testing data if we want to do inference with the trained model, and this inference step is done further in the module. A side-question could be: do we make predictions on a testing set with the trained model in *this* module?

tests/test_model_training.py
delhomer: Yes, my bad... This is fixed!

deeposlandia/kerastrain.py
garaud: It's a good question. The answer can wait. Let's use the code as is. We'll think about this if it's a problem or not.

garaud (approved):
looks good for me.

deeposlandia/kerastrain.py
delhomer: Ok, I agree.

Issue #52, opened 2018-04-18
test: test_semantic_segmentation_labelling sometimes fails


delhomer:
Without any visible reason, `test_semantic_segmentation_labelling` may fails to the following assertion:
```
assert list(bsum) == list(asum)
```
The problem seems to come from `np.bincount`, which does not produces `MAX-MIN` bins, if largest values are not into `a`.

PR #53, opened 2018-04-19
Labelling test updates

delhomer:
This PR refactors some test related to feature detection and semantic segmentation labelling.

Before there were three tests for each problem, that were quite redundant. Now, there are only two tests per problem:
- `test_problem_concise`, which tests the case where the array `a` contains exactly the values stored in `labels`
- `test_problem_sparse`, which tests the case where the array `a` values and `labels` are not the same (it corresponds to the case where there are some non-evaluated labels).

This PR fix issue #52.

### DIFF
--- a/tests/test_generators.py
+++ b/tests/test_generators.py
@@ -26,7 +26,7 @@ def test_feature_detection_labelling_concise():
                    [2, 1, 0, 0]]])
     labels = np.unique(a).tolist()
     MIN, MAX = np.amin(a), np.amax(a)
-    with pytest.raises(AttributeError):
+    with pytest.raises(ValueError):
         b = generator.feature_detection_labelling(a, ['0', '1', '2', '3'])
     b = generator.feature_detection_labelling(a, labels)
     assert b.shape == (a.shape[0], len(labels))
@@ -55,7 +55,7 @@ def test_feature_detection_labelling_sparse():
                    [1, 1, 1, 2, 1],
                    [1, 1, 2, 3, 2]]])
     labels = np.unique(a).tolist()[:-1]
-    with pytest.raises(AttributeError):
+    with pytest.raises(ValueError):
         b = generator.feature_detection_labelling(a, ['0', '1', '2'])
     b = generator.feature_detection_labelling(a, labels)
     assert len(labels) != np.amax(a) - np.amin(a) + 1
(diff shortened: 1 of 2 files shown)


tests/test_generators.py
```diff
@@ -8,58 +8,60 @@
...
-    """Test the value and the type of some label ids for feature detection one-hot encoding.
+    a = np.array([[[0, 0, 0, 2],
+                   [3, 3, 0, 2],
+                   [3, 3, 3, 0]],
+                  [[2, 2, 0, 0],
+                   [1, 2, 0, 0],
+                   [2, 1, 0, 0]]])
+    labels = np.unique(a).tolist()
+    MIN, MAX = np.amin(a), np.amax(a)
+    with pytest.raises(AttributeError):
```
garaud: I didn't see that for my first review when you updated some test, but this shouldn't be an `AttributeError` exception which should be raised. It's better if we raise a `ValueError` exception. We have to update the `feature_detection_labelling` function.

tests/test_generators.py
```diff
@@ -8,58 +8,60 @@
...
+    labels = np.unique(a).tolist()[:-1]
     with pytest.raises(AttributeError):
-        b = generator.feature_detection_labelling(labels, ['0', '3', '10'])
-
-    b = generator.feature_detection_labelling(labels, [0, 3, 10])
-    assert b.tolist() == [[True, True, True],
-                          [True, False, True]]
+        b = generator.feature_detection_labelling(a, ['0', '1', '2'])
+    b = generator.feature_detection_labelling(a, labels)
+    assert len(labels) != range(np.amin(a), np.amax(a))
```
garaud: Is it not strange to compare an integer to a generator here? This is always True anyway when you do `2 != range(0,3)`.

garaud (changes requested):
I'm not very familiar with the "concise" or "sparse" vocabulary. "sparse" makes me think to a sparsed array/matrix in linear algebra.

I'm not other propositions for renaming this for now. Let's keep that.

tests/test_generators.py
delhomer: The fact is that this portion of code raises an `AttributeError` ! If we replace it with a `ValueError`, the test does not pass any more.

tests/test_generators.py
delhomer: Oups, little forget. It is not `range(...)`, but `len(range(...)`. I will fix this.

garaud: Just (np.max(a) - np.min(a)) + 1 instead of len(range())

tests/test_generators.py
garaud: That's why i wrote "we have to update the `feature_detection_labelling` function" in that way.

tests/test_generators.py
delhomer: Ok, modified.

delhomer: By the way, it is `np.amax(a)-np.amin(a)`, instead of `np.amax(a)-np.amin(a)+1`.

delhomer: Hum, my bad, it is `np.amax(a)-np.amin(a)+1`. Modified.

garaud (approved):

PR #54, opened 2018-04-19
Test fixtures

delhomer:
Modify all test definition by introducing **test fixtures**. This means to define some transversal variables by the way of `pytest` methods with the appropriate decorator in a `conftest.py` module, and refer to them as parameters in each test that needs them.

Simple for integer or string variables, it becomes slightly more tricky for test data directory: we do not need to have each test data samples in the git repo. From now, raw Mapillary data sample is the only sample kept in the repo, other data samples are generated dynamically during testing process elsewhere on the system (actually, it is in `/tmp/pytest-somewhat/...`, corresponding samples and then deleted at the end of the working session) and do not need to be committed along with the project. `tests/data/` subdirectories have been cleaned accordingly.


garaud (approved):
looks good for me. Thanks!

PR #59, opened 2018-04-23
Model backup

delhomer:
Introduce backups in `kerastrain.py`. Training a model begins by getting old backups on the file system, if they exist. Then new backups are stored at the end of each training epoch if they improve the best score. For now the reference metric is the validation accuracy.

Additionally a new test has been added to control the model backup process.

### DIFF
--- a/deeposlandia/kerastrain.py
+++ b/deeposlandia/kerastrain.py
@@ -221,9 +221,10 @@ if __name__=='__main__':
             model.load_weights(checkpoint_complete_path)
             utils.logger.info(("Model weights have been recovered from {}"
                                "").format(checkpoint_complete_path))
-        utils.logger.info(("No available checkpoint for this configuration. "
-                           "The model will be trained from scratch."))
-        trained_model_epoch = 0
+        else:
+            utils.logger.info(("No available checkpoint for this configuration. "
+                               "The model will be trained from scratch."))
+            trained_model_epoch = 0
     else:
         utils.logger.info(("No available checkpoint for this configuration. "
                            "The model will be trained from scratch."))
--- a/deeposlandia/kerastrain.py
+++ b/deeposlandia/kerastrain.py
@@ -215,7 +215,7 @@ if __name__=='__main__':
     if os.path.isdir(checkpoint_path):
         checkpoints = os.listdir(checkpoint_path)
         if len(checkpoints) > 0:
-            model_checkpoint = checkpoints[-1]
+            model_checkpoint = max(checkpoints)
             trained_model_epoch = int(model_checkpoint[-5:-3])
             checkpoint_complete_path = os.path.join(checkpoint_path, model_checkpoint)
             model.load_weights(checkpoint_complete_path)
--- a/deeposlandia/kerastrain.py
+++ b/deeposlandia/kerastrain.py
@@ -230,10 +230,12 @@ if __name__=='__main__':
                            "The model will be trained from scratch."))
         trained_model_epoch = 0
 
-    checkpoint_filenames = os.path.join(folders["output"], "checkpoints",
-                                        instance_name, "checkpoint-epoch-{epoch:02d}.h5")
+    checkpoint_filename = os.path.join(folders["output"],
+                                       "checkpoints",
+                                       instance_name,
+                                       "checkpoint-epoch-{epoch:03d}.h5")
     checkpoints = callbacks.ModelCheckpoint(
-        checkpoint_filenames,
+        checkpoint_filename,
         monitor='val_acc',
         verbose=0,
         save_best_only=True,
--- a/deeposlandia/kerastrain.py
+++ b/deeposlandia/kerastrain.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from datetime import datetime
 
-from keras import backend
+from keras import backend, callbacks
 from keras.models import Model
 
 from deeposlandia import generator, utils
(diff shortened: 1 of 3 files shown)


deeposlandia/kerastrain.py
```diff
@@ -210,23 +210,56 @@ def add_training_arguments(parser):
     STEPS = args.nb_training_image // args.batch_size
     VAL_STEPS = args.nb_validation_image // args.batch_size
     TEST_STEPS = args.nb_testing_image // args.batch_size
+
+    checkpoint_path = os.path.join(folders["output"], "checkpoints", instance_name)
+    if os.path.isdir(checkpoint_path):
+        checkpoints = os.listdir(checkpoint_path)
+        if len(checkpoints) > 0:
+            model_checkpoint = max(checkpoints)
```
garaud, line 218: This a alphanum order. Will the selected file be always the most recent?

deeposlandia/kerastrain.py
```diff
@@ -210,23 +210,56 @@ def add_training_arguments(parser):
...
+        else:
+            utils.logger.info(("No available checkpoint for this configuration. "
+                               "The model will be trained from scratch."))
+            trained_model_epoch = 0
+    else:
+        utils.logger.info(("No available checkpoint for this configuration. "
+                           "The model will be trained from scratch."))
+        trained_model_epoch = 0
+
+    checkpoint_filenames = os.path.join(folders["output"], "checkpoints",
```
garaud: I prefer to have a variable name in singular.

deeposlandia/kerastrain.py
```diff
@@ -210,23 +210,56 @@ def add_training_arguments(parser):
...
+                                        instance_name, "checkpoint-epoch-{epoch:02d}.h5")
```
garaud: just a question: you assume with this `02d` that the number of epochs won't be higher than 1000?

tests/test_model_training.py
```diff
@@ -38,6 +38,31 @@ def test_model_training(shapes_image_size, shapes_sample, shapes_sample_config,
...
+    label_ids = [x['id'] for x in config['labels'] if x['is_evaluate']]
+
+    cnn = FeatureDetectionNetwork("test", image_size=shapes_image_size, nb_labels=len(label_ids))
+    model = Model(cnn.X, cnn.Y)
+    old_weights = model.get_weights()
+    checkpoint_path = os.path.join(str(shapes_temp_dir), "checkpoints")
+    if os.path.isdir(checkpoint_path):
+        checkpoints = os.listdir(checkpoint_path)
+        if len(checkpoints) > 0:
+            model_checkpoint = checkpoints[-1]
```
garaud: It's different from the previous `max(checkpoint_path)` in the implemented function `add_training_arguments`.

tests/test_model_training.py
```diff
@@ -38,6 +38,31 @@ def test_model_training(shapes_image_size, shapes_sample, shapes_sample_config,
...
+            trained_model_epoch = int(model_checkpoint[-5:-3])
+            checkpoint_complete_path = os.path.join(checkpoint_path, model_checkpoint)
+            model.load_weights(checkpoint_complete_path)
+    else:
+        trained_model_epoch = 0
+    new_weights = model.get_weights()
+    assert trained_model_epoch > 0
+    assert len(old_weights) == len(new_weights)
+    assert old_weights[0].shape == new_weights[0].shape
+    assert all(not np.allclose(old_weights[i], new_weights[i]) for i in range(len(new_weights)))
```
garaud: I wonder if we can use a numpy testing assert here, such as `np.testing.assert_almost_equal`

garaud (changes requested):
Just some comments.

deeposlandia/kerastrain.py
delhomer: As we define the checkpoint with the parameter `save_best_only=True`, and as the checkpoint names depend only on current training epoch, all the saved model are supposed to provide a better validation accuracy than the previous ones. In such a context, considering the most recent checkpoint guarantees recovering the best validation accuracy.

deeposlandia/kerastrain.py
delhomer: Yes, write. This is fixed.

deeposlandia/kerastrain.py
delhomer: It won't even be higher than 100. I assumed indeed than the number of epochs won't be larger than this amount, however this may be a strong assumption if we consider larger models, and/or if we decompose epoches in smaller sets.

I may modify `02d` in `03d` for a sake of care.

*As a complementary remark*, note that this name convention is directly related to your first comment about `max(filename)`.

tests/test_model_training.py
delhomer: Yes write, this is a small forget. This is supposed to be `max(checkpoints)`.

tests/test_model_training.py
delhomer: Why not, I did not know this possibility. I was wondering if the most accurate is really `assert_almost_equal`, as we handle integers. I see a `assert_allclose` which is close from the current used function `np.allclose`.

delhomer: Yark, we do not test if all weights are equivalent. The goal is the opposite one: to check if each new weight is different from each old weight. It seems that there is no convenient numpy test for such a task.

I tried a workaround with a `pytest` context manager, something like:
```
with pytest.raises(AssertionError):
    any(np.testing.assert_allclose(old_weights[i], new_weights[i]) for i in range(len(new_weights)))
```
However it is not clear for me that this snippet is really accurate (*e.g.* the test result is the same if I write `any(not ...)`).

garaud (approved):

Issue #60, opened 2018-04-24
training: unoperating learning rate


delhomer:
In the current state of the code, the learning rate is totally useless: the learning rate is considered as an parameter of the `ConvolutionalNeuralNetwork` class constructor, however it is not an attribute of this class.

In the new Keras-based API, the learning rate must be defined with the optimizer function, as follows (for instance):
```
opt = Adam(learning_rate=0.001, decay=0.0001)
model.compile(loss='binary_crossentropy', optimizer=opt, metrics=['acc'])
```

PR #61, opened 2018-04-24
training: fix the learning rate implementation

delhomer:
Before this PR, the learning rate was still implemented following the TensorFlow version of the code. The way the learning rate is considered is now updated: it is introduced as a parameter of the optimizer function, instead of being an attribute of the `ConvolutionalNeuralNetwork` class.

As a direct consequence of this treatment, we do not consider a parsed list of argument any more: a `-L` argument handles the starting learning rate, and a `-l` argument manages the learning rate decay rate.

This PR fixes the issue #60.


garaud (approved):
looks good for me

PR #62, opened 2018-04-27
Model inference

delhomer:
This PR introduces a new module that will make the inference easier in further developments (for instance, in a web app).

Until now, inference was done only after training a model. It was possible to do inference only, by passing 0 to `nb_epochs`, or alternatively a number smaller than the checkpoint training step, if a backup exists. This was quite unpractical, as the program arguments are training-focused.

Here we can predict labels on a given image by simply entering the following command (as an example):
```
python deeposlandia/inference.py -D dataset -i path_to_image/image_name.png
```

Additionally, the `prepare_folders` function has been splitted into three more precise functions:
- `prepare_input_folder`, which is called during dataset generation (*cf* `deeposlandia/datagen.py`).
- `prepare_prepro_folder` which is called during dataset generation and training process: such directories are filled during the former and the images that they contain are scanned during the latter.
- `prepare_output_folder` which is called when results are produced, either during training process (model backup creations) or during inference (model backup recovering).

### DIFF
--- a/deeposlandia/inference.py
+++ b/deeposlandia/inference.py
@@ -1,0 +1,213 @@
+
+"""Inference on deep learning model previously trained
+
+* Build the instance name
+
+* Load one (or more) image(s) from the file system:
+
+* Load a trained model starting from the instance name
+
+* Make label predictions on the test image(s)
+
+* Produce a result: for instance, only predicted labels
+
+Example of program call, that will infers labels on all files from
+̀path_to_images/shapes_00000.png` to `path_to_images/shapes_00009.png`::
+
+    python deeposlandia/inference.py -D shapes -i path_to_images/shapes_0000*.png
+
+"""
+
+import argparse
+import glob
+import numpy as np
+import os
+from PIL import Image
+import sys
+
+from keras.models import Model
+
+from deeposlandia import utils
+from deeposlandia.keras_feature_detection import FeatureDetectionNetwork
+from deeposlandia.keras_semantic_segmentation import SemanticSegmentationNetwork
+
+
+def add_program_arguments(parser):
+    """Add instance-specific arguments from the command line
+
+    Parameters
+    ----------
+    parser : argparse.ArgumentParser
+        Input parser
+
+    Returns
+    -------
+    argparse.ArgumentParser
+        Modified parser, with additional arguments
+    """
+    parser.add_argument('-D', '--dataset',
+                        required=True,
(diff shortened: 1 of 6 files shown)


deeposlandia/inference.py
```diff
@@ -0,0 +1,213 @@
...
+if __name__ == '__main__':
+
+    program_description = ("Infer labels on one (or more) image file(s) "
+                           "from a trained deep neural network")
+    parser = argparse.ArgumentParser(description=program_description)
+    parser = add_program_arguments(parser)
+    parser = add_instance_arguments(parser)
+    args = parser.parse_args()
+
+    image_paths = [item for sublist in [glob.glob(f) for f in args.image_paths]
```
garaud: You can turn these comprehension lists into generator expressions. Such as:

```python
(item for sublist in (glob.glob(f) for f in args.image_paths) for item in sublist)
```

Moreover, it's not clear for me, at least when I read the first time, that you want to flatten the list of image files. Can I propose something like:

```python
images_paths = (glob.glob(f) for f in args.image_paths)
# flatten the result of [[image1, image2], [image10, image11]]
image_paths = itertools.chain(*image_paths)
```

with the great package [itertools](https://docs.python.org/3/library/itertools.html)

deeposlandia/inference.py
```diff
@@ -0,0 +1,213 @@
...
+                     args.learning_rate, args.learning_rate_decay]
+    instance_name = utils.list_to_str(instance_args, "_")
+
+    input_folder = utils.prepare_input_folder(args.datapath, args.dataset)
+    prepro_folder = utils.prepare_preprocessed_folder(args.datapath,
+                                                      args.dataset,
+                                                      image_size,
+                                                      aggregate_value)
+
+    print( prepro_folder['training_config'] )
```
garaud: remove the print. Or replace it by a logging message.

garaud (changes requested):

deeposlandia/inference.py
delhomer: Useless 'print' statement, it has been removed.

deeposlandia/inference.py
delhomer: I agree, `itertools` is really convenient. However I'm not sure to want a iterator there.

The iterator is consumed in the first loop, when the `x_test` variable is built. But its items are not available any more in the last loop, when results are printed: how do we get the image filenames if such a structure is chosen?

garaud (approved):

Issue #63, opened 2018-04-27
Prepare 0.4 release


delhomer:
The transition between TensorFlow to Keras has meant a lot of changes in the code. Some new features are also available. It is time for releasing!

Steps to do before releasing:
- [x] merge PR #62 
- [x] clean the `deeposlandia` repository (part of PR #64 )
- [x] change the version number in `setup.py` (part of PR #64 )
- [x] read the code to catch some minor formatting issues, useless comments or outdated/missing docstrings (part of PR #64 )
- [x] update `README.md` (part of PR #64 )

delhomer:
@garaud if you think to additional steps, do not hesitate to amend the issue description.

Then I am not really sure to include the hyperparameter optimization in this release. What is your meaning about this question? Maybe we can let it for a 0.4.1... Your opinion is welcome here!

garaud:
We can add hyperparameter optimization in another relase.

The last digit of the semantic versioning, i.e. the `x.y.z` of `z` is often dedicated to fixed bugs. I prefer 0.5 if there are new features.

delhomer:
Ok, I'm fine with that. I modify the issue description accordingly.

PR #64, opened 2018-04-27
Code cleaning

delhomer:
Some cleaning operations to prepare the `0.4` release:
- version updatting in `setup.py`
- some useless modules deletion, some module renaming
- code reading (**in processing**)
- update `README.md`

Fixes the issue #63 .


setup.py
```diff
@@ -6,7 +6,7 @@
             version = line.strip().split('=')[1].strip(' \'"')
             break
     else:
-        version = '0.3.2'
+        version = '0.4'
```
garaud: maybe write also the version number in the `__init__.py` file. What to you think?

README.md
```diff
@@ -33,170 +28,167 @@ The picture below shows an example of image generated in this way:
...
-following program calls are supposed to be made from the `source` repository.
++ cv2
++ keras
++ h5py
++ logging
++ matplotlib
++ numpy
++ pandas
++ PIL
++ tensorflow
```
garaud, line 63: maybe mention the setup.py file with the dependencies and the requitements dev file if needed.

garaud (changes requested):
minor comments

setup.py
delhomer: Yes you're write, it seems to be a good practice. Should I modify the `setup.py` too, or should we keep the version in both scripts?

garaud: we can keep the version in both scripts. I'll look for another open source projects how they deal with that.

delhomer:
Some new commits have been added to fit your remarks. If everything is OK, we can merge, and release the new code version.

garaud (approved):

Issue #55, opened 2018-04-19
app: design a small app to visualize model inference results


delhomer:
In order to communicate the results of the neural network training, one possibility could be to design a small web application. It would allow visitors to use a trained model in order to get feature detection or semantic segmentation relatively to a sample of picture.

*To keep it simple:* let's imagine a web app that could load a trained model (`h5`) from a server, do inference on a single image (either loaded from the server, or chosen by the client, but the former seems to be a good first step), and return result onto the web page.

- [x] build a simple Flask web application structure
- [x] select a random Shapes picture from the server
- [x] create an image uploader in order to infer labels on client images
- [x] make the server infer on Shapes image labels for feature detection problem
- [x] make the server infer on Mapillary image labels for feature detection problem
- [x] make the server infer on Shapes image labels for semantic segmentation problem
- [x] make the server infer on Mapillary image labels for semantic segmentation problem

delhomer:
An app is provided by commit [496da3](496da3c97beb8bb71d2eeb653ef094817225faf0). It illustrates in a very simple manner the label inference for feature detection problem.

delhomer:
The app is now working. Further improvements will be focused on new dedicated issues.

PR #68, opened 2018-06-21
Webapp

delhomer:
This PR introduces a brand new way for visualizing deep learning predictions through a Flask web application.
The application is quite simple, as no real focus has been done on formatting web pages. However, it provides a first complete picture of the project results.

In order to make the Flask app happen, some symbolic link must be designed:
+ `deeposlandia/static/shapes` that points on shape image folder (the folder that contains a `images` and a `labels` subrepositories), for demo purpose.
+ `deeposlandia/static/mapillary` that points on mapillary image folder for demo purpose (*see* previous point).
+ `deeposlandia/static/predicted_images` that points to a server-side temporary folder to store predicted images (labelled versions of tested images).
+ `deeposlandia/static/sample_image` that points to `images` repository within the `deeposlandia` project, so as to display example images at web page first visits.

This PR fixes issue #55 .


deeposlandia/static/app.js
```diff
@@ -0,0 +1,19 @@
...
+var API_URL = PREFIX + '/api';
+
+// Build the URL with a BASE_URL/<city> suffix based from a DOM element with the
+// "city" dataset attribute, i.e. 'data-city'.
+function cityurl(dom_id) {
+  return API_URL + "/" + document.getElementById(dom_id).dataset.city;
+};
+
+// get the date before today in YYYY-MM-DD string format
+function getYesterday() {
```
garaud: Is this function mandatory? Maybe you don't neet it.

deeposlandia/static/app.js
```diff
@@ -0,0 +1,19 @@
+// Jitenshea
+
+// PREFIX should be defined in the settings.js file
+var API_URL = PREFIX + '/api';
+
+// Build the URL with a BASE_URL/<city> suffix based from a DOM element with the
+// "city" dataset attribute, i.e. 'data-city'.
+function cityurl(dom_id) {
```
garaud: Legacy copy/pasting from jitenshea I think. You can remove it.

deeposlandia/webapp.py
```diff
@@ -0,0 +1,153 @@
...
+PREDICT_FOLDER = os.path.join(PROJECT_FOLDER, 'predicted/')
+os.makedirs(UPLOAD_FOLDER, exist_ok=True)
+os.makedirs(PREDICT_FOLDER, exist_ok=True)
+
+daiquiri.setup(level=logging.INFO)
+logger = daiquiri.getLogger("deeposlandia-webapp")
+
+app = Flask(__name__)
+app.config['ERROR_404_HELP'] = False
+app.config['SWAGGER_UI_DOC_EXPANSION'] = 'list'
```
garaud: I think you don't use Swagger, you can remove it.

deeposlandia/webapp.py
```diff
@@ -0,0 +1,153 @@
...
+    if request.method == 'POST':
+        # check if the post request has the file part
+        if 'file' not in request.files:
+            print('No file part')
+            return redirect(request.url)
+        file = request.files['file']
+        # if user does not select file, browser also
+        # submit a empty part without filename
+        if file.filename == '':
+            print('No selected file')
```
garaud: remaining print

garaud (changes requested):
Great work! Congrats.

I think that the symbolic link part with the different image folders could be in a specific configuration file, read by the Flask app when it starts.

It's not very clear to me the name of the different routes: `demo_predictor`, `predictor`, `predictor_demo`. You don't know if it's a route which serves a HTML file, or a route which serves some JSON data. And I think you should add some docstring.

deeposlandia/static/app.js
delhomer: Yes, indeed. Actually the whole `app.js` is useless for `deeposlandia`. I remove it.

deeposlandia/static/app.js
delhomer: You're write, I remove the file from the project. :)

deeposlandia/webapp.py
delhomer: Done.

deeposlandia/webapp.py
delhomer: Actually this is a stackoverflow-oriented development issue, there... Maybe we can keep the log as the function behavior still have to be clarified. I replace `print` statements with `utils.logger.info`.

delhomer:
I agree regarding the symbolic link point. I will add a dedicated issue to keep it in our mind.
Additionally, I will add a new commit with docstrings for webapp.py methods.

garaud (approved):

PR #71, opened 2018-07-02
Shape labels

delhomer:
Update shape label management to get smarter and effective shape predictions for semantic segmentation, as well as a clearer ground-truth displaying.

What's new:
- Ground-truth labelled images are now displayed on the web app, with corresponding colored label names.
- Background label is now considered as a trainable label.
- Label colors are more contrasting (blue squares, red circles, green triangles and light grey background).
- The web app now comes with trained models that consider these four labels.


deeposlandia/dataset.py
```diff
@@ -309,13 +309,13 @@ class ShapeDataset(Dataset):
     """
 
     SQUARE = 0
-    SQUARE_COLOR = (0, 10, 10)
+    SQUARE_COLOR = (50, 50, 200)
```
garaud: maybe describe the color in comment: `# red`, etc.

tests/test_model_training.py
```diff
@@ -52,6 +52,7 @@ def test_model_backup_loading(shapes_image_size, shapes_sample_config, shapes_te
     model = Model(cnn.X, cnn.Y)
     old_weights = model.get_weights()
     checkpoint_path = os.path.join(str(shapes_temp_dir), "checkpoints")
+    print(checkpoint_path)
```
garaud: remaining print.

deeposlandia/generator.py
```diff
@@ -5,54 +5,87 @@
 
 from keras.preprocessing.image import ImageDataGenerator
 
+def recover_label_id(img, label_config):
```
garaud, line 8: We don't have to change now but I'm think it will be a great idea to store the labelled pixels by their label id instead of the related color. It won't be necessary to convert the RBG to a label id for the training part because the data will be stored as such.

deeposlandia/generator.py
```diff
@@ -5,54 +5,87 @@
...
+    if not (img.shape[1] == img.shape[2] and
+            (len(img.shape) == 3 or len(img.shape) == 4)):
+        raise ValueError(("Wrong image shape. Please provide batched images "
+                          "with equal width and height dimensions."))
+    label_ids = [item['id'] for item in label_config if item['is_evaluate']]
     if not all(isinstance(item, (int, np.uint8, np.uint32, np.uint64)) for item in label_ids):
         raise ValueError(("List of label IDs must contains "
                           "integers: {}").format(label_ids))
+    if len(img.shape) == 4 and img.shape[3] > 1:
+        img = recover_label_id(img, label_config);
```
garaud, line 55: what's happen when you have the case where the dimension is 3? e.g. (x, y, channel)

deeposlandia/generator.py
```diff
@@ -5,54 +5,87 @@
...
     img = img.squeeze().astype(np.uint8)
     one_hot_encoding = np.equal.outer(img, label_ids)
     return one_hot_encoding
 
 
-def feed_generator(datapath, gen_type, image_size, batch_size, seed=None):
+def feed_generator(datapath, gen_type, image_size, batch_size, col_mode, seed=None):
```
garaud: do you have a checking or a `ValueError` if the `col_mode` doesn't have the right value ?

garaud: I suggest `color_mode`. The `col_` in the variable name `col_mode` can mean "column", "color", etc.

deeposlandia/generator.py
```diff
@@ -119,15 +154,18 @@ def create_generator(dataset, model, datapath, image_size, batch_size, label_ids
     """
     if not dataset in ['shapes', 'mapillary']:
         raise ValueError("Wrong dataset name {}".format(dataset))
-    image_generator = feed_generator(datapath, "images", image_size, batch_size, seed)
+    image_generator = feed_generator(datapath, "images", image_size,
+                                     batch_size, 'rgb', seed)
     if inference:
         return image_generator
-    label_generator = feed_generator(datapath, "labels", image_size, batch_size, seed)
+    col_mode = 'grayscale' if dataset == 'mapillary' else 'rgb'
```
garaud: same comment about `col_mode`

deeposlandia/inference.py
```diff
@@ -150,7 +150,8 @@ def init_model(problem, instance_name, image_size, nb_labels, dropout, network):
 
 def predict(filenames, dataset, problem, datapath="./data", aggregate=False,
             name=None, network=None, batch_size=None, dropout=None,
-            learning_rate=None, learning_rate_decay=None, output_dir=None):
+            learning_rate=None, learning_rate_decay=None,
+            output_dir="/tmp/deeposlandia/predicted"):
```
garaud, line 154: Is the folder `/tmp/deeposlandia` created if it does not exist?

deeposlandia/webapp.py
```diff
@@ -238,14 +238,21 @@ def demo_image_selector():
...
-    labels = {item['category']: utils.RGBToHTMLColor(item['color'])
-              for item in config['labels']
-              if item['id'] in lbls}
+    if dataset == "shapes":
+        lbls = np.unique(server_label_image.reshape([-1, 3]), axis=0).tolist()
+        labels = {item['category']: utils.RGBToHTMLColor(item['color'])
+                  for item in config['labels']
+                  if item['color'] in lbls}
+    else:
+        lbls = np.unique(server_label_image).tolist()
```
garaud: what is the difference between `lbls` and `labels`? Maybe find another way to name `lbls`. Maybe you don't have to turn the numpy.array to a list for the `item['id'] in xxx` test.

garaud (changes requested):

deeposlandia/dataset.py
delhomer: Done.

tests/test_model_training.py
delhomer: Done.

deeposlandia/generator.py
delhomer: The point is we necessarily have to do such a conversion at least one time:
- if we store labels as RGB-pixels (the current option), the `recover_label_id` method is needed for deep learning model setting up;
- if we store labels with their IDs (the previous option), we have to design a similar method for handling labelled images each time we need to use them (standard file visualization, image displaying on the web application...).

deeposlandia/generator.py
delhomer: The code blow up to pieces... I will add an `else` statement so as to raise an exception and catch the case in a cleaner way.
Here the "valid" method use is when we have batched RGB-images (format = [batch_size, im_size, im_size, nb_channel], with nb_channels == 3). We could debate on the acceptability of unbatched images, however a better way of doing is surely to prevent that, as we handle only batched images during dataset generator creations.

delhomer: I read the code again, and I guess we can run the method with only 3 dimensions, to the strict condition that we have [batch_size, im_size, im_size] (see first `if` statement in the method). Actually that's how Mapillary dataset is managed.

The following PR should harmonize the labelled image format (Mapillary labelled images will be stored as RGB instead of IDs), and such an `if` statement will become obsolete.

deeposlandia/generator.py
delhomer: Done.

deeposlandia/generator.py
delhomer: Done.

deeposlandia/inference.py
delhomer: It is created in `webapp.py`, as this module calls `inference.predict(...)` several times. However, I can add a `os.makedirs` statement to guarantee the folder creation, for instance when the method is called in another context.

deeposlandia/webapp.py
delhomer: I modify the variable names:
- `lbls` -> `actual_labels` : labels that are actually on the test image
- `labels` -> `printed_labels` : same labels, but formatted in order to be displayed on the web app

Plus, I remove the `.tolist()` part, that seems useless, indeed.

delhomer:
The different comments have been taken into account. I'm ready to merge if it is OK for you, @garaud .

garaud (approved):

garaud:
Looks good to me. you can merge.

Issue #46, opened 2018-04-11
Refactor the label IDs when aggregating original labels


delhomer:
When aggregation is done, label IDs are currently chosen as the smallest IDs in each group. However it generates some issues when designing label generators.

One possible solution to overcome this could be to generate new integer IDs from `0` to `N-1` (`N` being the number of aggregated labels), to associate them to aggregated labels, and to keep a trace of original labels in the configuration file.

delhomer:
It seems that label aggregation does not work correctly. As an example, I generate an image with following label IDs: 2, 3, 4, 5, 6, 7, 8, 9, 10, 19, 23, 25, 52, whilst there are only 11 labels (from 0 to 10).

PR #72, opened 2018-07-09
Mapillary label aggregation

delhomer:
This PR fixes the way aggregated Mapillary datasets are generated.
First the `deeposlandia/aggregate_label.py` module has been simplified so as to generate a cleaner `config_aggregate.json`. Such a configuration file will be valid for each further-generated aggregated dataset. Labels are now aggregated regarding original dataset sub-classes (*i.e.* `animal`, `barrier`, `flat`, `human`, `marking`, `nature`, `object`, `rider`, `structure`, `support`, `traffic-sign`, `vehicle`, `void`). Some new characteristics are now extracted:
- `name`, which is related to label sub-class;
- `family`, that indicates the raw label macro-class;
- `contains_id`, which points to corresponding raw label IDs;
- `contains` gives the same information, but with raw label names separated with `...` in order to design a single string.

`deeposlandia/dataset.py` and `deeposlandia/webapp.py` have been refactored as well to consider this new aggregated dataset description. The modifications in the first file are related to Dataset creation and corresponding image backups on the file system (labelled images are now stored in an easier way: directly as RGB-formatted files). And the modification in the second file adapt the web application to new images produced on the server for the demo.

This PR fixes issue #46 .


delhomer:
The PR is not ready yet... The RGB-formatted labelled images works as in the Shape dataset case. As a consequence, we should merge the PR #71 first, and adapt the current PR accordingly (see `recover_label_id` method, in `generator.py`, which appears as fundamental here).

deeposlandia/aggregate_label.py
```diff
@@ -124,16 +59,18 @@ def aggregate_config(config, df):
...
-        for _, label in group.iterrows():
-            if label['id'] == label_id:
-                d = config['labels'][label_id]
-                d['id'] = label_id
-                d['group_name'] = label[KEY_AGG]
-                d['aggregate'] = []
-            else:
-                d['aggregate'].append((label['id'], label['name']))
+    nb_labels = len(df.new_label.unique())
+    intgen = (i for i in range(nb_labels))
```
garaud: I think you can replace it by `range()` which is already a generator in py3.

deeposlandia/aggregate_label.py
```diff
@@ -124,16 +59,18 @@ def aggregate_config(config, df):
...
-        label_id = int(group['id'].min())  # int conversion from int64 for JSON serialization
-        for _, label in group.iterrows():
-            if label['id'] == label_id:
-                d = config['labels'][label_id]
-                d['id'] = label_id
-                d['group_name'] = label[KEY_AGG]
-                d['aggregate'] = []
-            else:
-                d['aggregate'].append((label['id'], label['name']))
+    nb_labels = len(df.new_label.unique())
```
garaud: you can use `nunique()` to get the number of unique values.

deeposlandia/dataset.py
```diff
@@ -101,18 +102,24 @@ def add_label(self, label_id, label_name, color, is_evaluate,
...
+        contained_labels : list
+            List of raw labels aggregated by the current label
         """
         if label_id in self.label_info:
             utils.logger.error("Label {} already stored into the label set.".format(label_id))
             return None
         self.label_info.append({"name": label_name,
                                 "id": label_id,
-                                "category": (label_name if category is None
+                                "category": (label_name
```
garaud: the python ternary operator is often written in a single line

`val = rhs if predicate else other_value`

deeposlandia/dataset.py
```diff
@@ -256,8 +264,12 @@ def _preprocess(self, image_filename, output_dir, aggregate, labelling=True):
             if aggregate:
                 final_img_out = self.group_image_label(final_img_out)
 
-            labels = utils.mapillary_label_building(final_img_out, self.label_ids)
-            new_out_filename = os.path.join(output_dir, 'labels', label_filename.split('/')[-1])
+            labels = utils.mapillary_label_building(final_img_out,
+                                                    self.label_ids)
+            new_out_filename = os.path.join(output_dir, 'labels',
+                                            label_filename.split('/')[-1])
```
garaud: maybe `os.path.basename` can be used instead of `split('/')`. For instance, in Windows, this does not work.

deeposlandia/webapp.py
```diff
@@ -136,15 +136,16 @@ def demo_prediction():
     filename = request.args.get('img')
     filename = os.path.join("deeposlandia", filename[1:])
     dataset = request.args.get('dataset')
+    agg_value = dataset == "mapillary"
```
garaud, line 139: Does it mean that, for the webapp, the mapillary dataset is necessarily with aggregated labels?

deeposlandia/webapp.py
```diff
@@ -170,7 +171,7 @@ def prediction():
     utils.logger.info("file: {}, dataset: {}, model: {}".format(dataset, model,
                                                                 filename))
     predictions = predict([filename], "mapillary", "semantic_segmentation",
-                          aggregate=False, output_dir=PREDICT_FOLDER)
+                          aggregate=True, output_dir=PREDICT_FOLDER)
```
garaud, line 174: is it possible to use the  `agg_value` variable previously defined?

deeposlandia/webapp.py
```diff
@@ -230,22 +231,24 @@ def demo_image_selector():
     summarizes the label information for displyaing purpose
     """
     dataset = request.args.get('dataset')
-    server_folder = os.path.join("deeposlandia", "static", dataset, "images")
-    client_folder = os.path.join("/", "static", dataset, "images")
+    dataset_code = dataset + "_agg" if dataset == "mapillary" else dataset
+    server_folder = os.path.join("deeposlandia", "static", dataset_code, "images")
+    client_folder = os.path.join("/", "static", dataset_code, "images")
```
garaud: I think the STATIC Flask folder is defined in a Flask configuration variable. See http://flask.pocoo.org/docs/1.0/api/?highlight=static%20folder#flask.Flask.static_folder

garaud (changes requested):

deeposlandia/aggregate_label.py
delhomer: Yum, it does not work. However, I can do `intgen = iter(range(nb_labels))` instead. I found this article to illustrate the point: http://treyhunner.com/2018/02/python-range-is-not-an-iterator/

deeposlandia/aggregate_label.py
delhomer: Done.

deeposlandia/dataset.py
delhomer: I agree with you. What if the line is too long? Here I can define two variables in dedicated lines before the `append` statement to keep the code easily readable.

deeposlandia/dataset.py
delhomer: Yes, good point, I did not know the `basename` function, and missed the doc. I change that.

deeposlandia/webapp.py
delhomer: For now, yes (the result are better...). However we could imagine in a near future the management of both cases, with a checkbox on the Mapillary webpages!

deeposlandia/aggregate_label.py
garaud: You're right. I forgot this point. Thanks for sharing the article.

deeposlandia/dataset.py
garaud: As you wish.

deeposlandia/webapp.py
delhomer: The `agg_value` is defined in another method, starting from the dataset passed as a parameter in the app.

Here we know that we want to predict Mapillary labels (this method is relative to "try-it-yourself" part of the app, where clients may upload their own images), so it makes sense to consider a hard-coded aggregate value (True, as it was chosen to consider this case in the PR).

deeposlandia/webapp.py
garaud: Very good idea.

garaud:
I think you can delete the `WIP` part from the pull request title.

deeposlandia/webapp.py
garaud: OK, thanks.

deeposlandia/webapp.py
delhomer: Two modifications:
- os.path.join("deeposlandia", "static") -> app.static_folder
- os.path.join("/", "static") -> app.static_url_path

Issue #21, opened 2018-03-15
Implementation of the IoU (Intersection over Union) function for metrics and/or loss


garaud:
This function can be used as a metrics or loss function.

See:
* https://www.pyimagesearch.com/2016/11/07/intersection-over-union-iou-for-object-detection/
* http://angusg.com/writing/2016/12/28/optimizing-iou-semantic-segmentation.html

delhomer:
Fixed after PR #75 merging. I close it.

PR #75, opened 2018-07-20
Metric design

delhomer:
This PR aims at introducing new metrics to follow the training performance. We design Intersection-over-Union (IoU) as in https://en.wikipedia.org/wiki/Jaccard_index, and the Dice coefficient, as in https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient.

Two additional methods are added to consider metric-based losses, however there are not used for now. The used loss functions are still binary cross-entropy for feature detection and categorical cross-entropy for semantic segmentation, as they are very popular in the state-of-the-art... Further investigations could be done in the future on this particular question, however we can consider that these loss functions are satisfying.

All the metrics are logger in a `.csv` file in instance-related repositories, so as to keep track of each training process.

This PR fixes issue #21 .

### DIFF
--- a/deeposlandia/train.py
+++ b/deeposlandia/train.py
@@ -334,13 +334,16 @@ if __name__=='__main__':
                                         patience=10,
                                         verbose=1,
                                         mode='max')
+    csv_logger = callbacks.CSVLogger(os.path.join(output_folder,
+                                                  'training_metrics.csv'))
 
     hist = model.fit_generator(train_generator,
                                epochs=args.nb_epochs,
                                steps_per_epoch=STEPS,
                                validation_data=validation_generator,
                                validation_steps=VAL_STEPS,
-                               callbacks=[checkpoint, terminate_on_nan, earlystop],
+                               callbacks=[checkpoint, terminate_on_nan,
+                                          earlystop, csv_logger],
                                initial_epoch=trained_model_epoch)
     metrics = {"epoch": hist.epoch,
                "metrics": hist.history,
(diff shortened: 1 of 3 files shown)


deeposlandia/train.py
```diff
@@ -123,6 +123,96 @@ def add_training_arguments(parser):
...
+    Returns
+    -------
+    float
+        Intersection over Union value
+    """
+    actual = backend.flatten(actual)
+    predicted = backend.flatten(predicted)
+    intersection = backend.sum(actual * predicted)
+    union = backend.sum(actual) + backend.sum(predicted) - intersection
+    return 1. * intersection / union
```
garaud: Can you tell my why you multiply by 1. ?

deeposlandia/train.py
```diff
@@ -123,6 +123,96 @@ def add_training_arguments(parser):
...
+
+def iou_loss(actual, predicted):
```
garaud: I think if we have more loss/metrics functions, we will move them into a dedicated Python module.

garaud (changes requested):
Just a comment and one question. Looks good for me.

garaud (approved):

deeposlandia/train.py
delhomer: We want to get a float value, however `intersection` and `union` are tensors that contains integer values.

deeposlandia/train.py
delhomer: I agree. I have prepared a commit to put metric design methods in a `metrics.py` module.

PR #76, opened 2018-07-23
Readme

delhomer:
This PR completes `README.md` with information about the Flask web application and needed symbolic links. These links are defined into a `config.ini` file (that must be created at the project root on the model of `config.ini.sample`) and are now automatically created on the app loading.
The presence of `config.ini` file as well as those of a `symlink` section is verified before app launching.


garaud (approved):
looks good fr me.

PR #77, opened 2018-07-23
Network architectures

delhomer:
This PR brings some new network architectures for feature detection as well as for semantic segmentation. The goal of the PR is not to be exhaustive relatively to state-of-the-art architectures; hence only a few ones have been designed, merely for analyse purpose (compare two or three architectures, analyze the hardware behavior when training models on a larger network and so on...).
The implemented architectures in the feature detection case are:
- Inception (v4)
- ResNet
Regarding semantic segmentation, there is:
- U-net
- Dilated convolutions

### DIFF
--- a/deeposlandia/feature_detection.py
+++ b/deeposlandia/feature_detection.py
@@ -5,7 +5,7 @@
 import h5py
 
 import keras as K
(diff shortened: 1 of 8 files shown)


garaud (approved):
great work!

PR #80, opened 2018-08-16
Aerial image dataset

delhomer:
This PR introduces a new dataset: [AerialImageDataset](https://project.inria.fr/aerialimagelabeling/).

This new dataset is handled from the data preprocessing to the visualization on the web application. Hence:
- a new `AerialDataset` class that inheritates from `Dataset` has been added into `dataset.py` module;
- the dataset generation, the training, the hyperparameter tuning and the inference modules now consider this new alternative;
- the web application now contains a new webpage dedicated to the dataset, and all subsequent Flask app routes handle it.

This dataset contains very-large input images (5000*5000 pixels) that are cropped during the preprocessing step, so as to get smaller images that will be usable in a deep learning context with limited hardware capabilities (a small workstation with one GPU). As the neural network architectures requires divisible-per-32 image size, some image size adaptations are needed. New utilitary functions do this job in `utils.py`.


delhomer:
I've just added a test to verify the mapping between tile and image sizes, for this dataset. Additionally, the considered method docstring has been enriched with the expected output value for `raw_img_size==5000`.

I hope it is clearer with this new commit.

delhomer:
I merge in order to validate the integration of the dataset into the web application. In case of a subsequent untreated issue, we will open a new PR.

PR #96, opened 2019-01-03
Update CLIs

delhomer:
This PR harmonizes the different CLIs that compose the project, *i.e.* `datagen`, `train` and `paramoptim` programs. These programs have used different arguments until now for the image amounts, this point is fixed. Additionally, we modify the default number of images, to force user to specify a strictly positive quantity (default to 0 for training, validation and testing images).
Finally, the mentions of a testing dataset have been removed from `train.py` and `paramoptim.py` as they are obsolete now.


garaud (approved):
looks good for me

Issue #97, opened 2019-01-09
Prepare 0.5 release


delhomer:
Considering there has been some new big features since 0.4, we could prepare a new release. It will contains:
- two new datasets focusing on building footprint detection have been introduced
in the framework, namely [INRIA Aerial Image
dataset](https://project.inria.fr/aerialimagelabeling/) and [Open AI Tanzania
dataset](https://blog.werobotics.org/2018/08/06/welcome-to-the-open-ai-tanzania-challenge/).
- some state-of-the-art neural network architectures have been added to the framework, based on the Keras API;
- a bunch of jupyter notebooks that clarify the framework usage;
- a Flask web application that may showcase the project.

Remaining steps before the release:
- [x] manage issue #93 
- [x] merge [Tanzania data generation branch](https://github.com/Oslandia/deeposlandia/tree/tanzania_datagen)
- [x] manage issue #98

PR #88, opened 2018-12-11
WIP: Tanzania dataset integration

delhomer:
This PR introduces the Tanzania dataset into deeposlandia framework.

The main modifications arise as follows:
- a new module has been defined for defining Tanzania buildings (`deeposlandia/buildings.py`). It inherits from the `Dataset` class. The main part of the job is done here, as this dataset involves geo-referenced images. The project then needs to consider geographic tools like GDAL and geopandas.
- some minor modifications were needed to update docstrings and if-case in other modules (`datagen.py`, `generator.py`, `train.py`, `paramoptim.py`, `inference.py`).
- the web application has been updated (new webpage, reference on the index.html page).
- the README.md has been enriched with the new dataset description. Additionally, a new markdown document describes how to add new datasets, by listing all required operations.
- some new tests that consider Tanzania dataset have been designed, and a corresponding data sample has been added to `tests/data` folder.

The PR is still on processing, as it misses : 
- the training of a model (the README.md will have to be updated with a more convincing prediction example...)

PR #101, opened 2019-01-16
Tanzania datagen

delhomer:
This PR modifies the way Tanzania data are preprocessed, in order to produce smaller image patches. The main modification lies in the fact that `_preprocess(.)` method is splitted into two new methods (namely `_preprocess_for_training(.)` and `_preprocessed_for_inference(.)`.

Whilst the latter works as previously, the former picks random patches into original high-resolution image, by focusing on areas that contains buildings. This results in a more balanced dataset with a higher building rate.

One also modify the `extract_points_from_polygon(.)` function, by including into it the (x, y) shift.

Finally, the training process is slightly updated, by considering validation loss as the reference metrics for early stopping.

This PR is linked with #97 .

### DIFF
--- a/deeposlandia/datasets/tanzania.py
+++ b/deeposlandia/datasets/tanzania.py
@@ -68,6 +68,78 @@ class TanzaniaDataset(Dataset):
                        color=self.FOUNDATION_COLOR, is_evaluate=True)
 
 
+    def _generate_preprocessed_filenames(
+            self, image_filename, output_dir, x, y, suffix
+    ):
+        """Generate preprocessed image and label filenames on the file system,
+        starting from a raw image filename
+
+        Parameters
+        ----------
+        image_filename : str
+            Original image filename
+        output_dir : str
+            Output folder for preprocessed material
+        x : int
+            Extracted image west coordinates
+        y : int
+            Extracted image north coordinates
+        suffix : str
+            Preprocessed filename complement
+
+        Returns
+        -------
+        dict
+            Preprocessed image and corresponding label filenames
+        """
+        basename_decomp = os.path.splitext(os.path.basename(image_filename))
+        img_id_str = (str(self.image_size) + '_'
+                      + str(self.image_size) + '_'
+                      + str(x) + '_' + str(y) + "_" + suffix)
+        new_filename = basename_decomp[0] + '_' + img_id_str + ".png"
+        out_image_name = os.path.join(output_dir, 'images', new_filename)
+        out_label_name = out_image_name.replace("images", "labels")
+        return {"image": out_image_name, "labels": out_label_name}
+
+
+    def _serialize(
+            self, tile_image, labelled_image, label_dict,
+            image_filename, output_dir, x, y, suffix
+    ):
+        """Serialize a tiled image generated from an original high-resolution
+        raster as well as the labelled version of the tile
+
+        The method returns a dict that contains image-related file paths.
+
+        Parameters
+        ----------
+        tile_image : PIL.Image
+        labelled_image : PIL.Image
+        label_dict : dict
+        image_filename : str
+        output_dir : str
+        x : int
+        y : int
+
+        Returns
+        -------
+        dict
+            Information related to the serialized tile (file paths, encountered labels)
+        """
+        dirs = self._generate_preprocessed_filenames(
+            image_filename, output_dir, x, y, suffix
+        )
+        # gdal.Translate(dirs["image"], raster,
+        #                format="PNG",
+        #                srcWin=[x, y, self.image_size, self.image_size])
+        tile_image.save(dirs["image"])
+        labelled_image.save(dirs["labels"])
+        return {"raw_filename": image_filename,
+                "image_filename": dirs["image"],
+                "label_filename": dirs["labels"],
+                "labels": label_dict}
+
+
     def _preprocess_tile(self, x, y, image_filename, output_dir,
                          raster, labels=None):
         """Preprocess one single tile built from `image_filename`, with respect
@@ -94,51 +166,24 @@ class TanzaniaDataset(Dataset):
             Key/values with the filenames and label ids
 
         """
-        basename_decomp = os.path.splitext(
-            os.path.basename(image_filename))
-        img_id_str = (str(self.image_size) + '_'
-                      + str(self.image_size) + '_'
-                      + str(x) + '_' + str(y))
-        new_in_filename = (basename_decomp[0] + '_'
-                           + img_id_str + ".png")
-        new_in_path = os.path.join(output_dir, 'images',
-                                   new_in_filename)
-        gdal.Translate(new_in_path, raster,
+        dirs = self._generate_preprocessed_filenames(
+            image_filename, output_dir, x, y
+        )
+        gdal.Translate(dirs["image"], raster,
                        format="PNG",
                        srcWin=[x, y, self.image_size, self.image_size])
-        if not labels is None:
-            raster_features = get_image_features(raster)
-            tile_items = extract_tile_items(raster_features, labels,
-                                            x, y,
-                                            self.image_size,
-                                            self.image_size,
-                                            tile_srid=32737)
-            out_labelname = (new_in_path
-                             .replace("images", "labels"))
-            mask = self.load_mask(tile_items, raster_features, x, y)
-            label_dict = utils.build_labels(mask,
-                                            range(self.get_nb_labels()),
-                                            "tanzania")
-            labelled_image = utils.build_image_from_config(mask,
-                                                           self.labels)
-            labelled_image.save(out_labelname)
-            return {"raw_filename": image_filename,
-                    "image_filename": new_in_path,
-                    "label_filename": out_labelname,
-                    "labels": label_dict}
-        else:
-            return {"raw_filename": image_filename,
-                    "image_filename": new_in_path}
+        return {"raw_filename": image_filename,
+                "image_filename": dirs["image"]}
 
 
-    def _preprocess(self, image_filename, output_dir, labelling):
+    def _preprocess_for_inference(self, image_filename, output_dir):
         """Resize/crop then save the training & label images
 
         Parameters
         ----------
         image_filename : str
             Full path towards the image on the disk
-        datadir : str
+        output_dir : str
             Output path where preprocessed image must be saved
 
         Returns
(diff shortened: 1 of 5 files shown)


deeposlandia/datasets/tanzania.py
```diff
@@ -94,51 +164,24 @@ def _preprocess_tile(self, x, y, image_filename, output_dir,
...
-        else:
-            return {"raw_filename": image_filename,
-                    "image_filename": new_in_path}
+        return {"raw_filename": image_filename,
+                "image_filename": dirs["image"]}
 
 
-    def _preprocess(self, image_filename, output_dir, labelling):
+    def _preprocess_for_inference(self, image_filename, output_dir):
         """Resize/crop then save the training & label images
```
garaud, line 178: is the docstring up to date? according to the new name of the function, i.e. "inference".

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+        Returns
+        -------
+        dict
+            Key/values with the filenames and label ids
+        """
+        raster = gdal.Open(image_filename)
+        raw_img_width = raster.RasterXSize
+        raw_img_height = raster.RasterYSize
+        image_data = raster.ReadAsArray()
+        image_data = np.swapaxes(image_data, 0, 2)
```
garaud, line 227: I think you can write `image_data = np.swapaxes(raster.ReadAsArray(), 0, 2)`

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+        result_dicts = []
+        logger.info("Image filename: %s" % image_filename)
```
garaud, line 229: in a logger, you shouldn't use the `%` at the right side of the string.

`logger.info("Image filename: %s", image_filename)`

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+        logger.info("Raw image size: %s, %s" % (raw_img_width, raw_img_height))
+
+        label_filename = (image_filename
+                          .replace("images", "labels")
+                          .replace(".tif", ".geojson"))
+        labels = gpd.read_file(label_filename)
+        labels = labels.loc[~labels.geometry.isna(), ["condition", "geometry"]]
```
garaud, line 236: Maybe you can use `labels.geometry.notnull()` instead of `~labels.geometry.isna()`

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+        none_mask = [lc is None for lc in labels.condition]
```
garaud, line 237: Can `labels.condition.isnull()` work?

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
 
+    def _preprocess_for_training(self, image_filename, output_dir, nb_images):
+        """Resize/crop then save the training & label images
+
+        Parameters
+        ----------
+        image_filename : str
+            Full path towards the image on the disk
+        output_dir : str
+            Output path where preprocessed image must be saved
```
garaud, line 216: missing `nb_images` in the docstring

I think the prefix `nb` is very frenchy and not so used by anglophone.

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+                        tile_image, labelled_image, label_dict,
+                        image_filename, output_dir, x, y, "nw"
+                        )
+                    result_dicts.append(tiled_results)
+                    image_counter += 1
+                    empty_image_counter += 1
+            nb_attempts += 1
+        del raster
+        logger.info("Generate %s images after %s attempts."
+                    % (image_counter, nb_attempts))
```
garaud, line 307: same comment about the `%` in the logger message

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
-                                                     output_dir,
-                                                     raster, labels)
+                                                     output_dir, raster)
                 result_dicts.append(tile_results)
         del raster
         return result_dicts
 
 
+    def _preprocess_for_training(self, image_filename, output_dir, nb_images):
+        """Resize/crop then save the training & label images
```
garaud, line 209: maybe add some content to the docstring about the method you use: randomly pick a range of pixels of an image and try to carry out data augmentation.

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+                image_counter += 1
+                tile_image_se = tile_image_sw.transpose(Image.FLIP_LEFT_RIGHT)
+                labelled_image_se = labelled_image_sw.transpose(Image.FLIP_LEFT_RIGHT)
+                tiled_results_se = self._serialize(
+                    tile_image_se, labelled_image_se, label_dict,
+                    image_filename, output_dir, x, y, "se"
+                )
+                result_dicts.append(tiled_results_se)
+                image_counter += 1
+                del tile_image_se, tile_image_sw, tile_image_ne
```
garaud, line 293: is it necessary to delete these variables? Some memory issues when you didn't delete them?

deeposlandia/datasets/tanzania.py
```diff
@@ -195,15 +330,24 @@ class method genericity
         image_list = os.listdir(os.path.join(input_dir, "images"))
         image_list_longname = [os.path.join(input_dir, "images", l)
                                for l in image_list
-                               if not l.startswith('.')][:nb_images]
+                               if not l.startswith('.')]
+        nb_image_files = len(image_list_longname)
 
         logger.info("Getting %s images to preprocess..."
-                    % len(image_list_longname))
+                    % nb_image_files)
```
garaud, line 337: you can remove the `%` and replace it by a comma

deeposlandia/datasets/tanzania.py
```diff
@@ -195,15 +330,24 @@ class method genericity
...
         logger.info(image_list_longname)
-        with Pool() as p:
-            self.image_info = p.starmap(self._preprocess,
-                                        [(x, output_dir, labelling)
-                                         for x in image_list_longname])
+        if labelling:
+            nb_tile_per_image = int(nb_images/nb_image_files)
+            with Pool(processes=3) as p:
```
garaud, line 341: maybe the number of process can be parameterized? Either it's a parameter of the CLI or a new variable in a config file or a environment variable name.

deeposlandia/datasets/tanzania.py
```diff
@@ -242,36 +386,45 @@ def load_mask(self, buildings, raster_features, min_x, min_y):
...
+                                                 min_x, min_y)
             label_id = [label["id"] for label in self.labels
                         if label["name"] == row["condition"].lower()][0]
             mask = cv2.fillPoly(mask, [points], label_id)
         return mask
 
 
-    def extract_points_from_polygon(self, p, features):
-        """Extract points from a polygon
+def extract_points_from_polygon(p, features, min_x, min_y):
```
garaud, line 398: this function could be tested

deeposlandia/datasets/tanzania.py
```diff
@@ -150,29 +193,121 @@ def _preprocess(self, image_filename, output_dir, labelling):
...
+                tiled_results_se = self._serialize(
+                    tile_image_se, labelled_image_se, label_dict,
+                    image_filename, output_dir, x, y, "se"
+                )
+                result_dicts.append(tiled_results_se)
+                image_counter += 1
+                del tile_image_se, tile_image_sw, tile_image_ne
+                del labelled_image_se, labelled_image_sw, labelled_image_ne
+            else:
+                if empty_image_counter < 0.1 * nb_images:
```
garaud, line 296: maybe add a comment to clearly undertsand your intent

garaud (changes requested):

deeposlandia/datasets/tanzania.py
delhomer: Not at all, well seen. I'm preparing a commit to update all the involved docstrings...

deeposlandia/datasets/tanzania.py
delhomer: It is supposed not to work? Or is it a pep recommendation?

deeposlandia/datasets/tanzania.py
delhomer: Seems to be OK with this change. Code modified!

deeposlandia/datasets/tanzania.py
delhomer: Seems to be OK with this alternative function as well. Anyway I keep `labels.condition.notnull()`.

deeposlandia/datasets/tanzania.py
delhomer: Exactly, that was the point, before limiting the number of `Pool()` processes. I propose to remove these lines now.

delhomer: As a remark, there is a `del raster` some lines below. This deletion is mandatory as it implies a GDAL raster.

deeposlandia/datasets/tanzania.py
delhomer: Ok, I will put it in `config.ini` file (and update `config.ini.sample` accordingly), it makes sense to manage it depending on the running system.

deeposlandia/datasets/tanzania.py
delhomer: I have completed the method docstring to explain the generation process in more details.

deeposlandia/datasets/tanzania.py
delhomer: Done.

deeposlandia/datasets/tanzania.py
delhomer: Hum... That's a big point, as it implies to test geometric coordinates. To be fair, we should even test other functions of the module...
How about a next dedicated PR?

garaud: In a next dedicated PR it's a good idea. If you create the input data (simple data), the returned data should be tested.

deeposlandia/datasets/tanzania.py
garaud: Take a look at https://docs.python.org/3/howto/logging.html#logging-variable-data

delhomer: Ok the fact is I chose the wrong way of doing, as I wrote a "%" instead of a comma. I've applied your previous [comment](https://github.com/Oslandia/deeposlandia/pull/88#discussion_r242144585) too fast. Bad chance: this wrong way of doing does work, it would have been more simple if not. It is a pity that we let it as is after this previous PR.

*Proposition:* write a specific issue to address in a dedicated PR, as the problem is spread over **all** the project modules.

deeposlandia/datasets/tanzania.py
delhomer: Link with previous comment, I mark this discussion as resolved and keep the previous one opened.

deeposlandia/datasets/tanzania.py
delhomer: See #103

deeposlandia/datasets/tanzania.py
delhomer: See #104

garaud (approved):

PR #112, opened 2019-01-25
Dataset test updates

delhomer:
This PR introduces more flexibility in dataset population process.

From now, its multiprocessing behavior may be controlled, in the sense that the number of used processes becomes a `populate()` method parameter. The method is called in `deeposlandia/datagen.py`, where one retrieve the number of processes starting from the project configuration file.

Talking about the configuration file, it is now managed in `deeposlandia\__init__.py`, as it is used by the webapp and by the datasets.

Finally, a refactoring of unit tests contained into `tests/test_dataset.py` has been done: one distinguishes now between training and testing datasets when focusing on the Tanzania dataset, as PR #101 modifies its population process. Then the `population(.)` method is run without multiprocessing during the unit tests : some tests fail with it in a fairly mysterious manner...

### DIFF
--- a/deeposlandia/__init__.py
+++ b/deeposlandia/__init__.py
@@ -1,7 +1,10 @@
 """Deeposlandia package
 """
 
+from configparser import ConfigParser
 import logging
+import os
+import sys
 
 import daiquiri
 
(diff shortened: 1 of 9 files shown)


deeposlandia/__init__.py
```diff
@@ -13,3 +15,12 @@
...
 logger = daiquiri.getLogger("root")
+
+_DEEPOSL_CONFIG = os.getenv('DEEPOSL_CONFIG')
+_DEEPOSL_CONFIG = _DEEPOSL_CONFIG if _DEEPOSL_CONFIG is not None else "config.ini"
+config = ConfigParser()
+if os.path.isfile(_DEEPOSL_CONFIG):
+    config.read(_DEEPOSL_CONFIG)
+else:
+    logger.error("No file config.ini!")
+    sys.exit(1)
```
garaud, line 27: is the module `sys` imported somewhere?

garaud (changes requested):
just one comment. After that, looks good for me.

deeposlandia/__init__.py
delhomer: No, you're right, good catch! That's fixed!

PR #116, opened 2019-02-01
fix some dependencies

garaud:
If the tensorflow dependency is not the first one, other dependencies download
and install the numpy library, with an incompatible version with TensorFlow.

Also add 'rtree' used in geopandas.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,14 +17,15 @@ with open('README.md', 'rb') as f:
     readme = f.read().decode('utf-8')
 
 install_requires = [
+    'tensorflow<=1.10',
     'opencv-python<=3.4.0.12',
     'pillow<=5.0.0',
-    'tensorflow<=1.10',
     'keras<=2.2.4',
     'daiquiri<=1.5.0',
     'Flask<=1.0.2',
     'seaborn<=0.8.1',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -17,14 +17,15 @@ def find_version(*file_paths):
...
     'opencv-python<=3.4.0.12',
     'pillow<=5.0.0',
-    'tensorflow<=1.10',
     'keras<=2.2.4',
     'daiquiri<=1.5.0',
     'Flask<=1.0.2',
     'seaborn<=0.8.1',
-    'geopandas<=0.4.0']
+    'geopandas<=0.4.0',
+    'rtree<=0.8']
```
delhomer, line 28: Talking about the dependency order, should `rtree` be before `geopandas`? Or maybe it does not matter?

delhomer (approved):
Fine for me, thanks @garaud !

setup.py
garaud: I don't think so. `pip install geopandas` does not install `rtree`, so...

delhomer: Ok, that was more a curiosity question than an obstructing point. We will come back to it in case of problem later. OK for merging the PR!

PR #119, opened 2019-02-01
Create a module dedicated to geometry structure handling

### LINKED ISSUE #109: Vectorize building predictions
As an output of the semantic segmentation process, one gets rasterized information, *i.e.* a `numpy.array` that depicts either the label IDs or the the corresponding pixels. That results in a labelled version of the image.

In the case of building footprint detection and georeferenced images, it is possible to georeference the building themselves, by extracting vectorized information (and storing it as a `.geojson` file, for instance).

delhomer:
This PR aims at moving the geometry-related functions in a dedicated `geometries.py` module. Such functions were essentially in `datasets/tanzania.py`. Hence this last module will be dedicated only to the Tanzania dataset management.

`geometries.py` is going to be enriched with conversion functions (raster->vector, vector->raster), so as to fix #109 .

Then some unit tests cover the new defined functions.

:warning: the new code features will have to be documented in `markdown` files, as well as in a dedicated `jupyter` notebook in the `examples` folder.

### DIFF
--- a/deeposlandia/datasets/tanzania.py
+++ b/deeposlandia/datasets/tanzania.py
@@ -12,7 +12,6 @@ import os
 
 import cv2
 import daiquiri
-import fiona
 import geopandas as gpd
 import numpy as np
 from osgeo import gdal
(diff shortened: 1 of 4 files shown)


garaud (approved):
good work!

PR #121, opened 2019-03-08
train: append existing CSV when saving metric history fix #102

### LINKED ISSUE #102: Use existing training metric history when retrain an existing model
For now, if one uses an existing model and retrains it, the `training_metrics.csv` is overwritten, and previous metric history is lost.

One should manage the file writing process so as to append new lines to the existing files.

delhomer:
This PR allows to exploit existing training metric history. Now new training period metrics are appended to existing CSV files.

The PR fixes issue #102 .


garaud (approved):
good idea

Issue #124, opened 2019-03-08
Save a "best model" after training


delhomer:
The `training` process currently saves models in an instance-specific folder, while the `paramoptim` process saves models in instance-specific folders *and* a "best model" that is referred by `inference` (hence by the web app) and `postprocessing` processes.

In order to avoid the longer hyperparametrization process, one should maintain a best model even after simple training runs.

PR #125, opened 2019-03-13
Improve the deeposlandia CLI

delhomer:
This PR aims at factorizing the different command-lien arguments that exist for available programs. A `deeposlandia/__main__.py` contains now all the parsed argument material, and an entry points has been set up in `setup.py`.

In this simplification effort, some extra-work has been done:
- `train.py` and `paramoptim.py` have been merged by considering that "train" is a particular case of "paramoptim" where there is only one option for each parameter (doing hyperparameter optimization may now be done with `train.py`, by passing list parameters); this point fix issue #124 as a "best model" is saved after hyperparameter optimization process.
- The `README.md` file has been updated.
- The documentation has been improved; there is now one file per command in `./docs/`.


setup.py
```diff
@@ -64,4 +65,7 @@ def find_version(*file_paths):
     python_requires=">=3",
     install_requires=install_requires,
     packages=find_packages(),
+    entry_points = {
+        "console_scripts": ["deepo=deeposlandia.__main__:main"],
```
garaud, line 69: I suggest `deeposl` instead of `deepo`. It's not necessary.

README.md
```diff
@@ -74,9 +74,19 @@ For other OS, please visit the `GDAL` installation documentation.
 
 ## Running the code
 
-- [Data preprocessing](./docs/preprocessing.md)
+A command-line interface is proposed with 4 available actions (`datagen`,
+`train`, `infer` and `postprocess`), callable as follows:
+
+```
+deepo action --params
```
garaud: don't understand what `--params` means? I suggest:

`deepo [command] --options`

Maybe rename `actions` by `commands`

deeposlandia/__main__.py
```diff
@@ -0,0 +1,293 @@
...
+        "--image-size",
+        required=True,
+        type=int,
+        help="Image patch size, in pixels",
+    )
+    return parser
+
+
+def add_dataset_args(parser, available_datasets=AVAILABLE_DATASETS):
+    """
```
garaud: docstring is missing

deeposlandia/__main__.py
```diff
@@ -0,0 +1,293 @@
...
+        "-s",
+        "--image-size",
+        required=True,
+        type=int,
+        help="Image patch size, in pixels",
+    )
+    return parser
+
+
+def add_dataset_args(parser, available_datasets=AVAILABLE_DATASETS):
```
garaud, line 157: The default value `available_datasets` is a list. You shouldn't have this type for default function values. I suggest a tuple which is a immutable python type.

deeposlandia/__main__.py
```diff
@@ -0,0 +1,293 @@
...
+        "-D",
+        "--dataset",
+        required=True,
+        choices=available_datasets,
+        help="Dataset type",
+    )
+
+
+def add_nb_image_args(parser):
+    """
```
garaud: missing docstring

deeposlandia/__main__.py
```diff
@@ -0,0 +1,293 @@
...
+        "-v",
+        "--nb-validation-image",
+        type=int,
+        default=0,
+        help=("Number of validation images"),
+    )
+
+
+def add_training_args(parser):
+    """
```
garaud: missing docstring

garaud (changes requested):
Some comments.

Well done!

deeposlandia/__main__.py
delhomer: I changed the nature of `AVAILABLE_DATASETS` (and `GEOGRAPHIC DATASETS` and `AVAILABLE_MODELS` which have the same nature) to comply with your remark. Thanks!

README.md
delhomer: You're right, that will be better clearer. I modify it.

deeposlandia/__main__.py
delhomer: Fixed.

deeposlandia/__main__.py
delhomer: Fixed.

deeposlandia/__main__.py
delhomer: Fixed.

setup.py
delhomer: I prefer to keep the name short for a sake of usability (instead of aliasing it afterwards... that's already an alias, after all), if it ain't bother you I will let it as is.

garaud (approved):

PR #132, opened 2019-04-17
Typo fixed on string 18

kant:


delhomer:
@kant Thanks for the typo fixing!

PR #145, opened 2020-04-22
Fix random image choice in dataset preprocessing

obrix:
Fix random image choice by making sure we have a set of unique images and no repetition.

Also raise a warning if the number of images we want is higher that the number of file we have.
Limit the  number of images to preprocess to the number of available images.

### DIFF
--- a/deeposlandia/datasets/__init__.py
+++ b/deeposlandia/datasets/__init__.py
@@ -479,8 +479,14 @@ class GeoreferencedDataset(Dataset):
                         ],
                     )
         else:
+            if nb_image_files < nb_images:
+                logger.warning(
+                    "Asking to preprocess %s images, but only got %s files",
+                    nb_images, nb_image_files)
+                nb_images = nb_image_files
+                logger.warning("Preprocessing %s images..", nb_images)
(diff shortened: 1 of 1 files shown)


deeposlandia/datasets/__init__.py
```diff
@@ -479,8 +479,14 @@ def populate(
                         ],
                     )
         else:
+            if nb_image_files < nb_images:
+                logger.warning(
+                    "Asking to preprocess %s images, but only got %s files",
+                    nb_images, nb_image_files)
+                nb_images = nb_image_files
+                logger.warning("Preprocessing %s images..", nb_images)
```
delhomer, line 487: I guess we're going to preprocess `nb_image_files` images there, aren't we? I'm not sure the logged variable is the good one.

delhomer (approved):
Fine for me, thanks @obrix!

deeposlandia/datasets/__init__.py
delhomer: Ok, forget my comment, one does `nb_images = nb_iimage_files` one line before...

Issue #147, opened 2020-04-22
Confusion with the datagen CLI


delhomer:
The `datagen` command CLI is highly confusing, regarding the image quantities, as it means a tile quantity for training/validation on the first hand, and a raw image quantity for testing on the other hand.

The reason of this choice was justified as follows:
- we pick a high number of random bbox in the training/validation images in order to build a robust tile dataset ;
- whilst we only want to cover up the whole image in the testing case: we generate tiles following a regular grid defined by the tile size.

One should clarify the CLI, by harmonizing the `-t`, `-v` and `-T` definition: these arguments should refer to the **number of raw images** (like in the testing case). As a consequence, one could add an additional argument (*e.g.* `--nb-tiles-per-image`) to define the targetted amount of tiles obtained from a single image, for the training/validation cases.

Linked to #146

PR #148, opened 2020-04-24
Fix datagen for geographic datasets

delhomer:
The CLI arguments '-t', '-v' and '-T' represents amounts of raw images. While it does not really matter for 'shapes' and 'mapillary' datasets, it clarifies the way geographic datasets ('aerial' and 'tanzania') are considered, especially in training and validation scopes.

As a side effect, one needs an additional arguments, 'nb-tiles-per-image', that depicts the amount of tiles that will be drawn from the original raw images (training and validation context). Example: if '-t=10' and '--nb-tiles-per-image=10', one will pick 10 images within the original training set, and try to generate 10 tiles from each of them (same random process as before).

The documentation for preprocessing has been clarified in this way.

As a side topic, this PR updates some of the project dependencies; the identified vulnerabilities should be fixed (https://github.com/Oslandia/deeposlandia/network/alerts).

Linked issue: #147


obrix:
Works for me

obrix (approved):

